PR #3, opened 2026-06-10
fix: forward caller-supplied Authorization header in H1 and H2c paths

zacharyvmm:
## Problem

`HttpClient._do_request()` and `_send_h2c_via_upgrade()` silently drop the `Authorization` header when callers set it via `req.headers.set("Authorization", ...)`. The header never reaches the wire.

This blocks legitimate per-request auth patterns:
- **AWS Signature V4** — signature is computed per-request from body, headers, and timestamp
- **OAuth2 short-lived tokens** — refresh frequently, differ per endpoint
- **HMAC request signing** — digest covers request body, must be set just before sending

## Root cause

Two methods unconditionally filtered `authorization` from forwarded caller headers, expecting auth to always flow through `HttpClient._auth_header`:

1. `_do_request()` line 713 — the main HTTP/1.1 path
2. `_send_h2c_via_upgrade()` line 1854 — the H2C upgrade path

## Fix

**H1 path:** Remove `authorization` from the skip condition entirely. When `_auth_header` is empty (the default), callers must be able to set `Authorization` via `req.headers`.

**H2c path:** Only skip caller's `Authorization` when `_auth_header` is already set — matching the existing pattern in `_build_h2_request_headers()` (line 1619-1623).

Both paths remain compatible with `BasicAuth`/`BearerAuth`: those set `_auth_header` which is still emitted on line 707-708 (H1) and 1847-1848 (H2c).

## Testing

Reproduced against moto S3 mock server. Before the fix, any request with a per-request Authorization header returns 403 (header absent from wire). After the fix, the header reaches the server and requests succeed.

### DIFF
--- a/flare/http/client.mojo
+++ b/flare/http/client.mojo
@@ -710,7 +710,7 @@ struct HttpClient(Movable):
         # Forward caller-supplied headers (skip Host — already set)
         for i in range(extra_headers.len()):
             var k = extra_headers._keys[i]
-            if k.lower() != "host" and k.lower() != "authorization":
+            if k.lower() != "host":
                 wire += k + ": " + extra_headers._values[i] + "\r\n"
 
         if len(body) > 0:
--- a/flare/http/client.mojo
+++ b/flare/http/client.mojo
@@ -711,6 +711,13 @@ struct HttpClient(Movable):
         for i in range(extra_headers.len()):
             var k = extra_headers._keys[i]
             if k.lower() != "host":
+                # Only skip caller's Authorization when _auth_header is already set,
+                # matching the h2/h2c paths (see _build_h2_request_headers).
+                if (
+                    k.lower() == "authorization"
+                    and self._auth_header.byte_length() > 0
+                ):
+                    continue
                 wire += k + ": " + extra_headers._values[i] + "\r\n"
 
         if len(body) > 0:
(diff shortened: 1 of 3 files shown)


ehsanmok (changes requested):
Thanks! Could you add a test for this please?

ehsanmok:
Looks like CI failure is known flake.

zacharyvmm:
Updated PR pushed — addressed your two points:

1. Tightened the H1 change to match the h2c/h2 guard: only skip the caller's Authorization when `_auth_header` is already set. No more duplicate headers when both a client credential and a caller header are present.

2. Added three regression tests in tests/http/test_unified_http_client.mojo:
   - caller Authorization reaches the wire on H1 with no client credential
   - caller Authorization reaches the wire on H2c upgrade with no client credential
   - no duplicate Authorization when client credential and caller header are both set

Note: I could not run the tests locally because the flare test suite has pre-existing nightly incompatibilities (`UnsafePointer[UInt8](unsafe_from_address=0)` is no longer valid in 1.0.0b2 — needs `Optional[UnsafePointer]`). I fixed the one in flare/net/_libc.mojo that blocked the h1 path, but the other three (in ws/server.mojo, runtime/blocking.mojo, runtime/scheduler.mojo) are deeper refactors outside this PR's scope. CI will exercise them in their own environment.

The test names follow the existing convention (fork_server loopback against an in-process HttpServer, same as the bearer-auth tests above).

ehsanmok:
Sounds good. I'll verify. Will update b2 once released in stable.

ehsanmok:
Ok the CI is fixed. Had to force push (with lease) to rebase this.

ehsanmok (approved):

PR #4, opened 2026-06-21
fix(deps) update recipe.yaml to pull json 0.2.1 for b2 compatibility

bowyern:
flare builds against b2 now, but the recipe.yaml pulls in json package that won't compile due to gpu function changes


ehsanmok:
Shoot! great catch. Thanks.

PR #10, opened 2026-08-25
fix(reactor): unregister must not raise when EPOLL_CTL_DEL fails

winding-lines:
`Reactor.unregister` raises `_os_error("epoll_ctl DEL")` whenever the DEL syscall fails. Two failures there are routine rather than fatal:

- **EBADF** — whoever took ownership of the fd already closed it, so the kernel auto-removed it from the epoll set. This happens on any ownership hand-off; the case that surfaced it was a WebSocket upgrade passing the socket to a connection object that closes it when the handler returns.
- **ENOENT** — the fd was never armed.

Both describe exactly the post-state `unregister()` is trying to reach. Raising skips the `_registered.pop(fd)` at the end of the function and leaves a stale bookkeeping entry behind. Because the kernel hands out the lowest free descriptor, the very next `accept(2)` typically reuses that number, and `register` then rejects it as already-registered — so a perfectly good new connection is silently dropped.

The kqueue branch directly below already ignores its `EV_DELETE` result for this exact reason ("ENOENT on unregistered filter is not a fatal error"). This makes the epoll branch match, so the two backends present the same contract. The `Raises:` docstring is narrowed to match: the only remaining raise is the not-registered guard.

Adds `test_unregister_after_close_succeeds` and
`test_unregister_after_close_frees_fd_for_reuse` to tests/runtime/test_reactor.mojo. The second is the one that matters — it re-registers after the recycled-fd hand-back, which is where the stale entry actually bites.

Created with Claude.

### DIFF
--- a/flare/runtime/reactor.mojo
+++ b/flare/runtime/reactor.mojo
@@ -403,7 +403,7 @@ struct Reactor(Movable):
             fd: Previously-registered fd.
 
         Raises:
-            NetworkError: If ``fd`` is not registered or the syscall fails.
+            NetworkError: If ``fd`` is not registered.
         """
         if fd not in self._registered:
             raise NetworkError("fd " + String(fd) + " is not registered", 0)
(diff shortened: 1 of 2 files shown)


ehsanmok:
LGTM! thanks.

PR #11, opened 2026-08-25
feat(ws): expose the handshake Origin on WsConnection

winding-lines:
Browsers do not apply the same-origin policy to `ws://` connects and send no preflight, so any page a user visits can open a WebSocket to a server this process is running -- including one bound to loopback. The `Origin` header is the only signal in the handshake that says which site opened the socket, and today the server parses it and throws it away: `_read_upgrade_request` walks every header but returns only the `Sec-WebSocket-Key`, and `WsConnection` carries just `_stream` + `_peer`. A handler therefore has no way to reject a cross-site connect.

This threads it through:

- `_WsUpgradeRequest` -- a small carrier for the handshake fields the server keeps (`key` + `origin`). Both parsers now return it, so the byte-buffer and stream paths stay mirror images of each other as their docstrings promise.
- `WsConnection.origin`, populated on both `WsServer` accept paths (the struct-handler loop and `_handle_ws_connection`).

The value is retained verbatim, never validated: the library has no way to know which origins a deployment trusts, so policy stays with the handler. Absent `Origin` yields the empty string rather than an error -- non-browser clients routinely omit it, and rejecting those would break every existing non-browser client.

The new constructor parameter is defaulted, so all existing `WsConnection(stream^, peer)` call sites keep compiling. Callers of the two parsers move from `key` to `.key` (3 test sites, 1 example); the fuzz harness already discards the result and is unchanged.

Tests: three byte-level parser cases (present / absent / case-insensitive field name) plus an end-to-end loopback assert that the Origin reaches the handler on `WsConnection`.

Written with Claude

### DIFF
--- a/examples/basic/ws_server.mojo
+++ b/examples/basic/ws_server.mojo
@@ -152,10 +152,10 @@ def accept_and_upgrade(
     """
     var server_stream = srv._listener.accept()
     var peer = server_stream.peer_addr()
-    var key = _read_upgrade_request(server_stream)
-    var accept = _compute_accept_srv(key)
+    var upgrade = _read_upgrade_request(server_stream)
+    var accept = _compute_accept_srv(upgrade.key)
     _send_upgrade_response(server_stream, accept)
-    return WsConnection(server_stream^, peer)
+    return WsConnection(server_stream^, peer, upgrade.origin.copy())
 
 
 # ── Main ──────────────────────────────────────────────────────────────────────
--- a/flare/ws/server.mojo
+++ b/flare/ws/server.mojo
@@ -159,7 +159,22 @@ def _str_find_srv(s: String, sub: String) -> Int:
     return -1
 
 
-def _parse_ws_upgrade_bytes(data: Span[UInt8, _]) raises -> String:
+@fieldwise_init
+struct _WsUpgradeRequest(Movable):
+    """The handshake fields the server keeps from an Upgrade request.
+
+    Fields:
+        key: The ``Sec-WebSocket-Key`` header value (always present; the
+            parsers raise when it is missing).
+        origin: The ``Origin`` header value, or empty when the client
+            sent none. Non-browser clients routinely omit it.
+    """
+
+    var key: String
+    var origin: String
+
+
+def _parse_ws_upgrade_bytes(data: Span[UInt8, _]) raises -> _WsUpgradeRequest:
     """Parse an HTTP WebSocket Upgrade request from a byte buffer.
 
     Identical logic to ``_read_upgrade_request`` but reads from a
@@ -170,7 +185,7 @@ def _parse_ws_upgrade_bytes(data: Span[UInt8, _]) raises -> String:
         data: Raw HTTP/1.1 Upgrade request bytes.
 
     Returns:
-        The ``Sec-WebSocket-Key`` header value.
+        The retained handshake fields (``Sec-WebSocket-Key`` + ``Origin``).
 
     Raises:
         NetworkError: If the request is malformed or missing required headers.
@@ -193,6 +208,7 @@ def _parse_ws_upgrade_bytes(data: Span[UInt8, _]) raises -> String:
     _ = read_line(data, pos)
 
     var ws_key = String("")
+    var ws_origin = String("")
     var found_upgrade = False
     var found_connection = False
 
@@ -215,6 +231,11 @@ def _parse_ws_upgrade_bytes(data: Span[UInt8, _]) raises -> String:
         )
         if k == "sec-websocket-key":
             ws_key = v
+        elif k == "origin":
+            # Retained verbatim, not validated: the library has no way to
+            # know which origins a given deployment trusts. Handlers get
+            # it via ``WsConnection.origin`` and apply their own policy.
+            ws_origin = v
         elif k == "upgrade" and _lower_srv(v) == "websocket":
             found_upgrade = True
         elif k == "connection" and "upgrade" in _lower_srv(v):
@@ -229,10 +250,10 @@ def _parse_ws_upgrade_bytes(data: Span[UInt8, _]) raises -> String:
         raise NetworkError(
             "WebSocket upgrade request missing Sec-WebSocket-Key"
         )
-    return ws_key^
+    return _WsUpgradeRequest(ws_key^, ws_origin^)
 
 
-def _read_upgrade_request(mut stream: TcpStream) raises -> String:
+def _read_upgrade_request(mut stream: TcpStream) raises -> _WsUpgradeRequest:
     """Read an HTTP upgrade request and return the ``Sec-WebSocket-Key``.
 
     Reads until the blank line terminating HTTP headers.
@@ -241,7 +262,7 @@ def _read_upgrade_request(mut stream: TcpStream) raises -> String:
         stream: Accepted TCP stream.
 
     Returns:
-        The ``Sec-WebSocket-Key`` header value.
+        The retained handshake fields (``Sec-WebSocket-Key`` + ``Origin``).
 
     Raises:
         NetworkError: If the upgrade request is malformed or missing the key.
@@ -250,6 +271,7 @@ def _read_upgrade_request(mut stream: TcpStream) raises -> String:
     _ = _read_line_srv(stream)
 
     var ws_key = String("")
+    var ws_origin = String("")
     var found_upgrade = False
     var found_connection = False
 
@@ -272,6 +294,11 @@ def _read_upgrade_request(mut stream: TcpStream) raises -> String:
         )
         if k == "sec-websocket-key":
             ws_key = v
+        elif k == "origin":
+            # Retained verbatim, not validated: the library has no way to
+            # know which origins a given deployment trusts. Handlers get
+            # it via ``WsConnection.origin`` and apply their own policy.
+            ws_origin = v
         elif k == "upgrade" and _lower_srv(v) == "websocket":
             found_upgrade = True
         elif k == "connection" and "upgrade" in _lower_srv(v):
@@ -286,7 +313,7 @@ def _read_upgrade_request(mut stream: TcpStream) raises -> String:
         raise NetworkError(
             "WebSocket upgrade request missing Sec-WebSocket-Key"
         )
-    return ws_key^
+    return _WsUpgradeRequest(ws_key^, ws_origin^)
 
 
 def _send_upgrade_response(mut stream: TcpStream, accept: String) raises:
@@ -323,6 +350,7 @@ struct WsConnection(Movable):
     Fields:
         _stream: The underlying TCP stream.
         _peer: The remote socket address.
+        origin: The handshake ``Origin`` header (empty when absent).
 
     Example:
         ```mojo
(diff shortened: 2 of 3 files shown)


flare/ws/server.mojo
```diff
@@ -229,10 +250,10 @@ def _parse_ws_upgrade_bytes(data: Span[UInt8, _]) raises -> String:
         raise NetworkError(
             "WebSocket upgrade request missing Sec-WebSocket-Key"
         )
-    return ws_key^
+    return _WsUpgradeRequest(ws_key^, ws_origin^)
 
 
-def _read_upgrade_request(mut stream: TcpStream) raises -> String:
+def _read_upgrade_request(mut stream: TcpStream) raises -> _WsUpgradeRequest:
     """Read an HTTP upgrade request and return the ``Sec-WebSocket-Key``.
```
ehsanmok, line 257: The summary line got left behind when the return type changed: the `Returns:` block below it now says the right thing while this one still promises the key. Docstrings here describe what the function does now.

```suggestion
    """Read an HTTP upgrade request and return the retained fields.
```

flare/ws/server.mojo
```diff
@@ -215,6 +231,11 @@ def _parse_ws_upgrade_bytes(data: Span[UInt8, _]) raises -> String:
         )
         if k == "sec-websocket-key":
             ws_key = v
+        elif k == "origin":
+            # Retained verbatim, not validated: the library has no way to
+            # know which origins a given deployment trusts. Handlers get
+            # it via ``WsConnection.origin`` and apply their own policy.
+            ws_origin = v
```
ehsanmok, line 238: This is rationale, and it's already in the commit message and in the `WsConnection.origin` docstring at `flare/ws/server.mojo:369-391`, both at more length than three lines allow. Comments in this repo are for constraints the code can't express, and `ws_origin = v` doesn't carry one. The same three lines sit in the stream parser at `flare/ws/server.mojo:297-301`.

```suggestion
        elif k == "origin":
            ws_origin = v
```

flare/ws/server.mojo
```diff
@@ -272,6 +294,11 @@ def _read_upgrade_request(mut stream: TcpStream) raises -> String:
...
```
ehsanmok, line 301: This is the second copy of the rationale comment, twin of the one in `_parse_ws_upgrade_bytes` at `flare/ws/server.mojo:234-238`. Same request: drop it and keep the assignment. The commit message and the `WsConnection.origin` docstring already carry the reasoning, and this repo keeps comments to constraints the code can't express.

```suggestion
        elif k == "origin":
            ws_origin = v
```

flare/ws/server.mojo
```diff
@@ -337,10 +365,40 @@ struct WsConnection(Movable):
...
+    policy of its own. Browsers do NOT apply the same-origin policy to
+    ``ws://`` connects and send no preflight, so **any** page a user
+    visits can open a WebSocket to a server this process is running --
+    including one bound to loopback. ``Origin`` is the only signal in
+    the handshake that says which site opened the socket.
+
+    Trust it only as far as browsers make it trustworthy: a browser sets
+    it and script cannot forge it, but a non-browser client sends
+    whatever it likes (or nothing at all). It authenticates the *page*,
+    never the user.
```
ehsanmok, line 382: The paragraph above sends a reader off to compare this against an allow-list, so it should say where that comparison isn't what it looks like. Three cases, read off the loop at `flare/ws/server.mojo:232-238`: a present-but-empty `Origin:` yields `""` exactly like an absent header, since `v` is stripped and assigned unconditionally; a repeated `Origin` keeps the last one seen; and a browser sends the literal `"null"` for an opaque origin, which matches no allow-list entry and isn't the same as absent. That last one is where people slip.

```suggestion
    Trust it only as far as browsers make it trustworthy: a browser sets
    it and script cannot forge it, but a non-browser client sends
    whatever it likes (or nothing at all). It authenticates the *page*,
    never the user.

    Three cases an allow-list check has to handle: an absent header and
    a present-but-empty ``Origin:`` both yield ``""``; a repeated
    ``Origin`` header keeps the last value seen; and a browser sends the
    literal ``"null"`` for an opaque origin (sandboxed iframe, ``data:``
    URL, some redirect chains), which matches no real origin and is not
    the same as absent.
```

flare/ws/server.mojo
```diff
@@ -337,10 +365,40 @@ struct WsConnection(Movable):
...
+
+    Example:
+        ```mojo
+        def on_connect(mut conn: WsConnection) raises:
+            if conn.origin != "https://app.example.com":
+                conn.close(WsCloseCode.POLICY_VIOLATION)
```
ehsanmok, line 388: Worth a clause here saying what this example does and doesn't do. Both accept paths send the 101 before they build the connection, at `flare/ws/server.mojo:700` and `:731`, so by the time a handler reads `origin` the socket is already upgraded and this `close` is a policy close on a live WebSocket rather than a refused handshake. That's the correct move for a server shaped this way and I'm not asking you to reshape it, I just don't want a reader taking the example as the upgrade being rejected. Refusing before the 101 needs a pre-upgrade hook on `WsServer`, which I'll take as a follow-up.

examples/basic/ws_server.mojo
```diff
@@ -152,10 +152,10 @@ def accept_and_upgrade(
     """
     var server_stream = srv._listener.accept()
     var peer = server_stream.peer_addr()
-    var key = _read_upgrade_request(server_stream)
-    var accept = _compute_accept_srv(key)
+    var upgrade = _read_upgrade_request(server_stream)
+    var accept = _compute_accept_srv(upgrade.key)
     _send_upgrade_response(server_stream, accept)
-    return WsConnection(server_stream^, peer)
+    return WsConnection(server_stream^, peer, upgrade.origin.copy())
```
ehsanmok, line 158: `send_upgrade_request` in this file sends no `Origin` (the request is built at `examples/basic/ws_server.mojo:57-67`), so this always copies an empty string and nothing in the example ever shows the new field. The "Demonstrates" list at the top of the module is unchanged too. Since the example is here to be read, I'd rather it showed the thing: add an `Origin:` line to that request and print `conn1.origin` in step 2 next to the 101 status line. Failing that, the constructor default already covers this call site and it can go back to two arguments. Minor either way, and `upgrade` is dead on the next line, so a field transfer would save the copy if the compiler takes one - I haven't tried it.

ehsanmok (approved):
Thanks for this - it's a clean, self-contained change and the reasoning sits in the commit message where it belongs. I pulled the branch and ran it before reading the diff: `tests/ws/test_ws.mojo` is 37/37 green, `examples/basic/ws_server.mojo` compiles, `mojo format` is clean on the three changed files. This is also the one open PR that still merges into `main` without conflicts, so there's nothing to rebase here; the only merge chore is item 1. Approving. Items 2 through 4 are docstring and comment work - say which of them you want and I'll take the rest.

| what I ran | result |
|---|---|
| `tests/ws/test_ws.mojo` | 37 tests run: 37 passed, 0 failed, 0 skipped (35.8 s) |
| `test_upgrade_parse_captures_origin` | PASS (0.005 s) |
| `test_upgrade_parse_origin_absent_is_empty` | PASS (0.004 s) |
| `test_upgrade_parse_origin_header_name_case_insensitive` | PASS (0.003 s) |
| `test_ws_connection_carries_handshake_origin` | PASS (0.373 s) |
| build `examples/basic/ws_server.mojo` | compiles |
| `mojo format` | clean on the 3 changed files |

## 1. Run `pixi run tests-gen` after you merge `main`

Your base predates the test-aggregation change, so here is the note I'll point the other three PRs in this batch at. Tests are now compiled per area into one binary under `tests/_agg`, driven by `pixi run tests` -> `tools/run_test_aggregates.sh`. The generated aggregate imports every module-level `def test_*` by name, matched by the `TEST_DEF` regex at `tools/gen_test_aggregates.py:95` across every file the glob at `:146` finds. Your four new functions aren't in that generated file, so under `pixi run tests` they wouldn't run at all, and CI calls the generator with `--check` first (`tools/run_test_aggregates.sh:66`), which fails on a stale aggregate rather than quietly dropping the coverage. The test file itself merges fine; it's the generated side that goes out of date, and one `pixi run tests-gen` with the regenerated `tests/_agg` committed alongside fixes it.

I also diffed your parsers against the ones on `main` today. The header loop is untouched there, so the textual merge should compile as well as merge.

## 2. Origin semantics the docstring leaves open

I agree with the stance: retain verbatim, validate nothing, leave policy to the handler. The docstring tells a reader to compare `conn.origin` against an allow-list, so it should also say where that comparison isn't what it looks like. Three cases, all read off the loop at `flare/ws/server.mojo:232-238`:

- A present-but-empty `Origin:` gives `""`, the same value as an absent header, since `v` is stripped and assigned unconditionally. A handler that reads `""` as "no browser here, non-browser client, let it through" treats a hostile empty header the same way. I don't think that's a hole on its own, since a non-browser client can send whatever origin it likes anyway and your docstring says so, but I'd rather the ambiguity were written down than discovered.
- A repeated `Origin` keeps the last one seen. Browsers send one; a request arriving through a rewriting intermediary may not.
- A browser sends the literal `"null"` for an opaque origin: sandboxed iframe, `data:` URL, some redirect chains. It matches no allow-list entry, it isn't the same as absent, and it's the one people get wrong.

Two sentences in the field docstring cover all three; suggestion inline. Two more byte-level cases would pin the first two in code, and `_upgrade_bytes` in your test file already makes them cheap:

```mojo
def test_upgrade_parse_origin_empty_value_is_empty() raises:
    """A present-but-empty Origin is indistinguishable from an absent one."""
    from flare.ws.server import _parse_ws_upgrade_bytes

    var raw = _upgrade_bytes("Origin:\r\n")
    assert_equal(_parse_ws_upgrade_bytes(Span[UInt8, _](raw)).origin, "")


def test_upgrade_parse_duplicate_origin_keeps_last() raises:
    """A repeated Origin header keeps the last value seen."""
    from flare.ws.server import _parse_ws_upgrade_bytes

    var raw = _upgrade_bytes(
        "Origin: https://a.test\r\nOrigin: https://b.test\r\n"
    )
    var up = _parse_ws_upgrade_bytes(Span[UInt8, _](raw))
    assert_equal(up.origin, "https://b.test")
```

## 3. The handshake is already done before a handler sees the origin

Both accept paths send the 101 and then construct the connection: `flare/ws/server.mojo:700` in the struct-handler loop, `:731` in `_handle_ws_connection`. So the docstring example, which closes with `POLICY_VIOLATION`, is closing a socket that has already been upgraded. That's the correct move for a server shaped this way, and I'm not asking you to reshape it here. I'd just like one clause in the docstring so nobody reads the example as the upgrade being refused. Refusing before the 101 needs a pre-upgrade hook on `WsServer`, which is mine to write.

## 4. Two comments and one stale docstring line

The rationale comment above `ws_origin = v` appears in both parsers and says what the struct docstring at `:369-391` and the commit message already say at greater length. Comments here are for constraints the code can't express. Suggestions inline for both copies.

`_read_upgrade_request`'s summary line still reads "return the `Sec-WebSocket-Key`" while the `Returns:` block below it was updated. Suggestion inline.

## On the symmetry claim

It holds. I read the two header loops side by side and they're identical: same `_lower_srv` on the field name, same `.strip()` on both sides of the colon, same last-wins assignment, same order in the `elif` chain, same two raises at the bottom. Header-name case, whitespace trimming, duplicate `Origin` and the empty-value case therefore behave the same on both paths, which is what the commit message claims.

That identity now runs about forty lines, and this PR adds six more to each copy. `_read_line_srv` already hands the stream path one line at a time, so `_read_upgrade_request` could buffer to the blank line and hand off to `_parse_ws_upgrade_bytes`, leaving a single loop. Not this PR's job; I'll open it as a follow-up.

## On the tests

The three byte-level cases and the end-to-end one all do what their docstrings say, and I confirmed the four names in the run log. None of them fails on the old code in the assertion sense - they can't compile against it, since `up.key` doesn't exist on a `String` return. For a new field that's the expected shape, so no complaint.

Coverage is lopsided between the two parsers, which is worth knowing rather than fixing now. The byte parser gets present, absent and case-insensitive. The stream parser gets one present case, end to end. Its absent path does run in the three loopback tests, which thread `upgrade.origin.copy()` through, but nothing asserts `""` there. With the loops identical today the risk is small, and it stops being small exactly when they drift, which is the argument for folding them together.

## Accuracy of the commit message

"The new constructor parameter is defaulted, so all existing `WsConnection(stream^, peer)` call sites keep compiling" is true, and no call site in the tree exercises it: all seven moved to the three-argument form. The default is there for out-of-tree callers. Fine by me, I'd only phrase it as what it is. The rest of the message checks out against the diff: both accept paths populated (`:701`, `:732`), three test sites and one example moved to `.key`, and the fuzz harness at `fuzz/fuzz_ws_server.mojo:47` does discard the result and needs no change.

A public `origin` field rather than an accessor matches `Request.peer` at `flare/http/request.mojo:53`, so no objection there either.

## What I'll take

- a pre-upgrade hook on `WsServer` so a handler can refuse before the 101 goes out
- folding `_read_upgrade_request` onto `_parse_ws_upgrade_bytes` so there's one header loop

From you:

- `pixi run tests-gen` once you've merged `main`
- the docstring wording in items 2, 3 and 4, or a word that you'd rather I did it

ehsanmok:
Merged as-is. Maintainer edits are off on the fork, so I took items 2 through 4 myself in d282574 on main (docstring wording, the two rationale comments, the example now sends and prints an Origin) and regenerated the test aggregate so the four new tests run under `pixi run tests`. Thanks again for the find.

PR #14, opened 2026-08-28
feat(client): bound body reads, not just connect

winding-lines:
`HttpClient`'s `timeout_ms` is documented as "Connect + read timeout", but it only ever reaches `connect_timeout` / `_connect_with_fallback`. Once the connection is up, a read can block forever.

That is what a half-open connection looks like from the client: the request goes out, the network drops, the peer never learns it should FIN, and `recv` waits for bytes that will never arrive. Reproduced against a bound `TcpListener` that never calls `accept` -- the kernel completes the handshake from the backlog, so `connect` succeeds and the request is written -- a default client was still blocked after 25s with no timeout in sight.

`with_read_timeout(ms)` sets `SO_RCVTIMEO` on the socket, so a stalled read raises `Timeout: recv` instead. It follows the existing move-in / move-out builder shape (`with_pool`, `with_prefer_http3`). `0` is the default and preserves today's behaviour exactly, so nothing changes for existing callers.

The bound is per-read inactivity, not a deadline on the whole response: a large body that keeps arriving never trips it, and a stalled one trips it after `ms` of silence. `PostHocDeadline` remains the whole-request deadline.

Arming happens at dial time, at all ten sites that create a connection -- direct and proxied, cleartext and TLS, one-shot, chunked-upload, streaming-download, and both pooled-h1 dials (fresh and stale-retry). Because the option rides on the socket, a connection that goes into the keep-alive pool keeps its timeout for its whole lifetime; the CONNECT tunnel's own proxy-response read is bounded too. `TlsStream` grows a `set_recv_timeout` that delegates to the TCP socket it owns.

Not covered: HTTP/3 over QUIC, which does its own UDP-side timing and has no equivalent knob here.

`timeout_ms`'s docstring is corrected to say what it actually does.

Tests: a silent peer now fails with `Timeout` rather than hanging (asserted on the error text, not merely on "something raised" -- the same shape passes trivially on a ConnectionRefused if the listener is let go early); a generous timeout does not disturb a normal round-trip, a pooled round-trip, or the default client.

Created with Claude


flare/http/client.mojo
```diff
@@ -305,6 +305,13 @@ struct HttpClient(Movable):
     :meth:`with_cookies` opts in. When enabled, :meth:`_send_once`
     captures ``Set-Cookie`` response headers and replays them as a
     ``Cookie`` request header. Freed in :meth:`__deinit__`."""
+    var _read_timeout_ms: Int
+    """Opt-in read timeout applied to every connection this client
+    dials, via ``SO_RCVTIMEO`` on the underlying socket. ``0``
```
ehsanmok: `_dial_http3` at client.mojo:846 is not armed, so "every connection this client dials" overshoots. QUIC sets its own recv timeout at flare/quic/client.mojo:565, which is a good reason to leave it alone, and you say as much in the commit message. It belongs here too, where someone reading the field will look for it.

```suggestion
    """Opt-in read timeout applied to every TCP and TLS connection
    this client dials, via ``SO_RCVTIMEO`` on the underlying socket.
    HTTP/3 is excluded: QUIC does its own UDP-side timing. ``0``
```

flare/http/client.mojo
```diff
@@ -596,6 +609,57 @@ struct HttpClient(Movable):
...
+        half-open connection looks like to the client -- the network
+        drops after the request goes out, the peer never learns it
+        should FIN, and ``recv`` waits for bytes that will never
+        arrive. ``ms`` sets ``SO_RCVTIMEO`` on the socket, so a read
+        that stalls that long raises ``Timeout`` instead.
+
+        The bound is per-read inactivity, not a deadline on the whole
+        response: a large body that keeps arriving never trips it, and
+        a stalled one trips it after ``ms`` of silence. Use
+        ``PostHocDeadline`` for a whole-request deadline.
```
ehsanmok: `PostHocDeadline` is a server-side `Handler` wrapper (flare/http/reliability.mojo:258). It runs its inner handler to completion and substitutes a 504 afterwards, and its own docstring is explicit that it does not preempt. It cannot bound an `HttpClient` request, and it would not cut a hung read short even where it does apply. There is no client-side whole-request deadline in the tree today, so the honest version is to say that:

```suggestion
        The bound is per-read inactivity, not a deadline on the whole
        response: a large body that keeps arriving never trips it, and
        a stalled one trips it after ``ms`` of silence. The client has
        no whole-request deadline.
```

flare/http/client.mojo
```diff
@@ -596,6 +609,57 @@ struct HttpClient(Movable):
...
+
+        Set it before the first request. The timeout is armed when a
+        connection is dialled and rides on the socket, so it covers
+        pooled connections for their whole lifetime -- but connections
+        already pooled keep whatever was armed when they were dialled.
```
ehsanmok: Both halves of this are off. Cleartext pooled connections do get re-armed on checkout: `_send_h1_pooled` rewraps the pooled fd into a `TcpStream` at client.mojo:2132 and this call lands at 2135 on both branches. Pooled TLS connections do not, because `_send_h1_tls_pooled_once` at client.mojo:1113 never arms. And since `_arm_read_timeout` is a no-op at zero (client.mojo:654), it never clears an option that is already set, so `with_read_timeout(0)` after a nonzero value leaves the timeout on every pooled cleartext fd - which the Args line below says it removes.

```suggestion
        Set it before the first request. The timeout rides on the
        socket, so a connection keeps it for its whole lifetime.
        Cleartext pooled fds are re-armed on checkout; pooled TLS
        connections keep whatever was armed when they were dialled,
        and lowering ``ms`` to ``0`` clears neither.
```

flare/http/client.mojo
```diff
@@ -1014,6 +1079,7 @@ struct HttpClient(Movable):
             stream = TlsStream.connect_timeout(
                 u.host, u.port, tls_cfg^, self._timeout_ms
             )
+        self._arm_read_timeout(stream)
```
ehsanmok, line 1090: Arming here is too late for the direct branch. `TlsStream.connect_timeout` (flare/tls/stream.mojo:637) calls `_connect_with_fallback` at line 673 and then a blocking `_do_ssl_connect` at 685 with no deadline on it, so a peer that completes the TCP handshake and then goes quiet hangs inside the handshake and never reaches this line. That docstring already promises "Maximum milliseconds for TCP + TLS handshake combined" at line 649, which it has never done.

The tunnel branch three lines up is fine, because `_connect_tunnel` armed the socket at client.mojo:1363 before `connect_over_tcp` wrapped it. So HTTPS through a proxy is bounded and HTTPS direct is not. Arming before `_do_ssl_connect` inside `connect_timeout` closes both, and it is what would let the new test cover TLS at all - pointed at an `https://` never-accept listener today, the test hangs instead of failing.

flare/http/client.mojo
```diff
@@ -2095,6 +2168,7 @@ struct HttpClient(Movable):
             raise NetworkError("HTTP/1.1 pooled request failed")
 
         var fresh = _connect_with_fallback(host, port, self._timeout_ms)
+        self._arm_read_timeout(fresh)
```
ehsanmok, line 2179: Not asking you to fix this here, but it is worth knowing about. A read timeout on a pooled fd is indistinguishable from a stale keep-alive: the read sits inside the `try` that ends at client.mojo:2160-2161, and any raise there falls through to this retry, so the whole request goes out a second time with the `POST` body attached. Against a live-but-silent peer that means two sends and two full timeouts before anything surfaces. The TLS pool has the same shape at client.mojo:1140-1152.

`_is_idempotent` at client.mojo:131 already exists for the retry policy, so gating this on a stale-connection signature rather than on any raise is the natural fix. I will open an issue for it.

flare/tls/stream.mojo
```diff
@@ -769,6 +769,22 @@ struct TlsStream(Movable, Readable):
                 raise NetworkError("TLS EOF before buffer full")
             received += n
 
+    def set_recv_timeout(self, ms: Int) raises:
+        """Bound blocking reads on the underlying TCP socket.
+
+        ``SO_RCVTIMEO`` applies to the socket, not to the TLS record
+        layer, so a stalled ``read`` raises ``Timeout`` after ``ms``
+        milliseconds without data on the wire -- the case a half-open
+        connection produces when the network drops mid-response.
```
ehsanmok, line 855: The blocking client path reads through `flare_ssl_read` (flare/tls/ffi/openssl_wrapper.cpp:180), not the `_ex` variant that classifies `SSL_get_error`. With `SO_RCVTIMEO` armed, `recv` gives `EAGAIN`, `SSL_read` returns -1 with `SSL_ERROR_WANT_READ`, and the wrapper falls into its `n < 0` branch and calls `capture_openssl_errors()` on an empty error queue. `TlsStream.read` at line 751-752 then raises `NetworkError("TLS read error: ")` with nothing after the colon, so `Timeout` never appears on this path and the new test's `err.find("Timeout")` assertion would fail against an `https://` URL.

What I would rather have is `read` mapping the retryable case to `Timeout("recv")`, since `flare_ssl_read_ex` already returns `FLARE_SSL_IO_WANT_READ` for it and the client would then read the same on both wires. Failing that, the docstring has to describe what happens:

```suggestion
        ``SO_RCVTIMEO`` applies to the socket, not to the TLS record
        layer. OpenSSL reports the expiry as ``SSL_ERROR_WANT_READ``,
        which :meth:`read` surfaces as ``NetworkError`` with an empty
        reason, so a stalled TLS read does not raise ``Timeout`` the
        way a stalled cleartext read does.
```

tests/http/test_client_read_timeout.mojo
```diff
@@ -0,0 +1,136 @@
...
+from flare.tcp import TcpListener
+from flare.testing import fork_server, kill_forked_server
+
+
+def _hello(req: Request) raises -> Response:
+    return ok("read-timeout-hello")
+
+
+def _url(port: Int) -> String:
+    return String("http://127.0.0.1:") + String(port) + String("/")
```
ehsanmok, line 55: The never-accept listener is the right instrument and it travels: `TcpListener.bind` listens with a backlog of 128 (flare/tcp/listener.mojo:87), so the kernel completes the handshake from the accept queue on macOS and Linux alike, and 300 ms against a peer that never sends a byte has nothing to race. What is missing is a `https://` sibling to this helper, which is the case where the behaviour actually differs - the error is a `NetworkError` with an empty reason there, not a `Timeout`.

Adding it needs item 2 fixed first, since against a never-accept listener the TLS path hangs in the handshake today rather than failing. Once it does fail, a TLS case here is the thing that keeps the docstring honest.

ehsanmok (changes requested):
Thanks for this - the hole is real and you diagnosed it correctly. `timeout_ms` has claimed "Connect + read" since it was written and only ever reached `connect`, and a half-open peer does park a client indefinitely. I ran the branch locally before reviewing and the cleartext half holds up exactly as you describe. The TLS half does not, and that is what I need closed before merge: on `https://` a stalled read surfaces as `NetworkError` with an empty reason, and a direct HTTPS dial against your own never-accept listener still hangs inside the handshake. Items 1 and 2. Items 3 and 4 are docstrings that send the reader somewhere the code does not go.

On the rebase: this branch predates the 2026-09-11 migration on `main` and conflicts only in `pixi.toml`. Run `pixi run tests-gen` after rebasing so `tests/http/test_client_read_timeout.mojo` joins `tests/_agg`, rather than hand-writing it into the `tests` chain you appended to, which is gone; see my note on #11 for what that runner does.

## What I ran

Branch head 2d45be9, Mojo 1.0.0, macOS arm64.

| entry | result | wall |
|---|---|---|
| `tests/http/test_client_read_timeout.mojo` | pass, prints `4 passed` | 23 s |
| `tests/http/test_h1_client_pool.mojo` | pass | 22 s |
| `tests/http/test_tls_client_pool.mojo` | pass, `4 tests run: 4 passed, 0 failed` | 4 s |
| `tests/http/test_client_proxy.mojo` | pass | 15 s |

`mojo format` is clean on all three changed `.mojo` files. I did not run anything on Linux, and I have no stalled-TLS reproduction, so items 1 and 2 come from reading the code and the FFI wrapper rather than from a run.

## The ten sites do check out

I walked every dial in `flare/http/client.mojo` on this branch. Ten `_arm_read_timeout` calls, and every TCP or TLS dial in the file sits behind one:

- `_connect_h2` happy-eyeballs probe, client.mojo:961, armed at 964
- `_send_h2_or_h1_tls`, tunnel at 1077 and direct at 1079, armed at 1082
- `_send_chunked_tcp`, 1219, armed at 1220
- `_send_chunked_tls`, 1253, armed at 1256
- `_connect_tunnel`, 1362, armed at 1363, which is what bounds the CONNECT response read you mention
- streaming download cleartext, tunnel 1446 and direct 1448, armed at 1449
- streaming download TLS, tunnel 1512 and direct 1514, armed at 1517
- `_do_request` cleartext, tunnel 2052 and direct 2054, armed at 2057, covering h2c prior-knowledge and the h2c Upgrade dance both
- `_send_h1_pooled`, fresh dial at 2134 and the rewrapped pooled fd at 2132, armed at 2135
- `_send_h1_pooled` stale retry, 2170, armed at 2171

`_send_h1_tls_pooled_once` at client.mojo:1113 receives an already-open stream, so there is nothing to arm there. `_dial_http3` at client.mojo:846 is the one dial left out, which matches your commit message if not the code. See the minor items.

## 1. A stalled TLS read raises the wrong error, and nothing covers it

The blocking client path reads through `flare_ssl_read` (flare/tls/ffi/openssl_wrapper.cpp:180), not the `_ex` variant that classifies `SSL_get_error`. With `SO_RCVTIMEO` armed, `recv` gives `EAGAIN`, `SSL_read` returns -1 with `SSL_ERROR_WANT_READ`, and the wrapper takes its `n < 0` branch and calls `capture_openssl_errors()` against an empty error queue. `TlsStream.read` at flare/tls/stream.mojo:750-752 turns that into `NetworkError("TLS read error: ")` with nothing after the colon. Your PR text says "a stalled read raises `Timeout: recv`", and the new `set_recv_timeout` docstring at flare/tls/stream.mojo:775-778 states it as fact; both hold on cleartext only. Point your own `err.find("Timeout")` assertion at an `https://` URL and it fails.

Two ways out. The cheap one is the docstring correction I left inline. The one I would rather have is classifying the retryable case: `flare_ssl_read_ex` already returns `FLARE_SSL_IO_WANT_READ` for exactly this condition, so `TlsStream.read` can map it to `Timeout("recv")` and the client then reads the same on both wires. Either is fine with me. The docstring as written is not.

Worth a look while you are in there: `read` returning 0 is a clean EOF to `_read_all_tls` (flare/http/_client/parse.mojo:36-39), so if `SSL_read` ever returns 0 rather than -1 on the timeout, a stalled body truncates silently into a parsed `Response`. I could not produce that, and -1 is what I expect from a socket BIO, but the gap between a hard error and silent truncation here is one return value wide.

## 2. A direct HTTPS dial is still unbounded through the handshake

`_arm_read_timeout` runs after `TlsStream.connect_timeout` returns (client.mojo:1079-1082), and `connect_timeout` calls `_connect_with_fallback` at flare/tls/stream.mojo:673 and then a blocking `_do_ssl_connect` at 685 with no deadline on it at all. Its own docstring promises "Maximum milliseconds for TCP + TLS handshake combined" at line 649, which was already wrong before this PR. So a peer that completes the TCP handshake and then says nothing hangs the client forever on `https://`, which is the case your first paragraph opens with and the case your never-accept listener produces.

The proxied branch is bounded, because `_connect_tunnel` armed the tunnel socket at client.mojo:1363 before `connect_over_tcp` wrapped it. HTTPS through a proxy times out, HTTPS direct does not, and I do not want to ship that line. Arming before `_do_ssl_connect` inside `connect_timeout`, or threading the recv timeout into it, closes both, and it is what makes the TLS test from item 1 writable at all.

## 3. `PostHocDeadline` is a server middleware

The docstring at client.mojo:627-628 sends client users to `PostHocDeadline` for a whole-request deadline. That is a `Handler` wrapper on the server side (flare/http/reliability.mojo:258): it runs its inner handler to completion and substitutes a 504 afterwards, and its own docstring is explicit that it does not preempt. It cannot bound an `HttpClient` call. There is no client-side whole-request deadline in the tree today, and I would rather the docstring said so. Suggestion inline.

## 4. The pool sentence is off in both directions

Cleartext pooled connections are re-armed on checkout, because `_send_h1_pooled` rewraps the pooled fd into a `TcpStream` at client.mojo:2132 and `_arm_read_timeout` lands at 2135 on both branches. Pooled TLS connections are never re-armed, since `_send_h1_tls_pooled_once` at client.mojo:1113 has no arm call. And `_arm_read_timeout` is a no-op at zero (client.mojo:654), so it never clears an option that is already set: `with_read_timeout(0)` after a nonzero value leaves the timeout on every pooled cleartext fd, which the Args line at 636-637 says it removes. Suggestion inline.

## Minor

- The h3 exclusion lives in your commit message and not in the code. The field docstring at client.mojo:309 says "every connection this client dials", and `_dial_http3` at 846 is not one of them. Suggestion inline.
- Both `_arm_read_timeout` docstrings (client.mojo:652 and 658) say "freshly dialled", and the TCP one is also called on a rewrapped pooled fd. The behaviour is right, the word is not.
- A read timeout on a pooled connection is indistinguishable from a stale keep-alive, so the request goes out a second time on a fresh dial, `POST` included. Inline at client.mojo:2171; I am not asking you to fix it here.
- `SO_SNDTIMEO` is untouched, so a peer that stops reading during a `send_chunked` upload still blocks forever in `write_all`. The option name is honest about its scope, so I am not asking for it in this PR. `TcpStream.set_send_timeout` already exists at flare/tcp/stream.mojo:636 and raises `Timeout("send")`, so a `with_write_timeout` follow-up is short.
- `raise Timeout("recv")` at flare/tcp/stream.mojo:442 drops the `ms`, so the message reads `Timeout: recv` where the type can render `Timeout: recv (300 ms)`. Pre-existing and optional, and it would let your assertion be tighter than a substring search.

## On the test itself

The never-accept trick is sound and portable: `TcpListener.bind` listens with a backlog of 128 (flare/tcp/listener.mojo:87), so the kernel finishes the handshake out of the accept queue on macOS and Linux alike. The margins are safe too - 300 ms against a peer that never sends a byte has nothing to race, and 30 s on the live cases is generous. Asserting on the message rather than on "something raised" is the right call, and the `idle_count() == 1` check in the pooled case earns its place. The gap is TLS, which is where the behaviour actually differs.

## What I'll take

The send-side timeout and the pooled-retry misclassification are mine, and I'll open issues for both so they don't sit on you.

From you: items 1 through 4, and a TLS case in the new test once item 2 is in. Two of those are one-line docstring edits and item 2 is the only one with real work in it.

If you would rather scope this PR to cleartext, that works for me too - say so, make the docstrings and the commit message say cleartext, and I will merge it and take TLS myself. The one thing I cannot merge is a docstring promising TLS coverage that is not there.

winding-lines:
Thanks — items 1 and 2 were both real, and chasing item 1 turned up a third thing I would have shipped as a regression. Rebased onto `d282574`, one commit, `pixi run tests-gen` run.

The `pixi.toml` conflict resolved to main's side wholesale; the `tests` chain I appended to is gone and `tests/_agg/agg_http.mojo` now carries the six tests.

## 2. The handshake bound

`connect_timeout` arms `SO_RCVTIMEO` for the duration of `_do_ssl_connect` and clears it once the handshake completes, so a returned stream reads unbounded exactly as before and the caller still arms its own read timeout if it wants one. `timeout_ms` is now a bound on each phase rather than one shared budget, and the docstring says that instead of "TCP + TLS handshake combined".

To make the expiry a timeout rather than a `TlsHandshakeError("")` I added `flare_ssl_connect_ex` to the wrapper: same SNI + `SSL_set1_host` + `SSL_connect` as `flare_ssl_connect` (both now go through one `flare_ssl_connect_raw`), but it calls `SSL_get_error` **before** `capture_openssl_errors()` drains the queue it reads, and returns the `FLARE_SSL_IO_*` sentinel. Against your never-accept listener over `https://`:

```
Timeout: tls_handshake (300 ms)
```

## 1. The stalled TLS read

Took the classification rather than the docstring. Reproduced your diagnosis exactly first — the pre-fix error is `NetworkError: TLS read error: ` with nothing after the colon.

Not through `flare_ssl_read_ex`, though, and this is the part worth your attention: switching `read` to it broke `test_https_alpn_h2_is_served`, deterministically, 3 runs out of 3. `flare_ssl_read_ex` calls a peer that vanished without `close_notify` FATAL — OpenSSL 3.x reports that as `SSL_ERROR_SSL` / "unexpected eof while reading", which `flare_ssl_io_classify`'s `SSL_ERROR_SYSCALL && ret == 0` branch does not catch — where plain `flare_ssl_read` returns it as a 0. The h2-over-TLS client drains to EOF and depends on that 0.

So `read` goes through a new `flare_ssl_read_blocking` instead: bytes, 0 for **any** end of stream (`flare_ssl_read`'s contract, unchanged), WANT_READ / WANT_WRITE split out, FATAL otherwise. On a blocking fd those two can only mean the socket's own `SO_RCVTIMEO` / `SO_SNDTIMEO` expired. `_do_ssl_read` is gone; `read` was its only caller.

That also closes the one-return-value gap you flagged: a timeout can no longer arrive as a 0-byte read and truncate a stalled body into a parsed `Response`.

Worth saying out loud since it is adjacent: the reactor's `flare_ssl_read_ex` still calls an unclean EOF fatal. I left it alone — it is a different path with different callers — but if that is wrong it is wrong today too, independent of this PR.

## 3, 4 and the minors

Took your suggestions verbatim: `PostHocDeadline` out of the client docstring with "The client has no whole-request deadline" in its place, the pool sentence corrected in both directions, the h3 exclusion moved into the field docstring at :309. Both `_arm_read_timeout` docstrings lost "freshly dialled" — the TCP one now says "a fresh dial, or a pooled fd rewrapped on checkout", and the TLS one says pooled TLS connections are not re-armed.

Left for you as you said: the pooled-retry misclassification and `SO_SNDTIMEO`. I also left `Timeout("recv")` dropping the `ms` at `tcp/stream.mojo:442` — the handshake raise does carry it, so the two read differently for now.

## Tests

Six now, four unchanged. The two new ones:

- `test_tls_handshake_timeout_fires_on_silent_peer` — your `https://` sibling. Uses the literal IP so there is no `::1` first attempt to race; nothing on that path reaches certificate verification.
- `test_read_timeout_fires_on_stalled_tls_response` — past the handshake, against a real `bind_tls` server whose handler sleeps 3 s against a 300 ms bound. `localhost` there, because that one does verify and OpenSSL matches `SSL_set1_host` against the DNS SANs.

Both assert on the message.

## What I ran

macOS arm64, Mojo 1.0.0 (ed45d567), at `b501005`.

| entry | result |
|---|---|
| `tests/http/test_client_read_timeout.mojo` | 6 passed (3 consecutive runs) |
| `tests/http/test_h1_client_pool.mojo` | 4 passed |
| `tests/http/test_tls_client_pool.mojo` | 4 run, 4 passed |
| `tests/http/test_client_proxy.mojo` | 1 passed |
| `tests/tls/test_tls.mojo` | 16 run, 16 passed |
| `tests/tls/test_tls_resume.mojo` | 3 run, 3 passed |
| `tests/tls/test_https_reactor.mojo` | 8 run, 8 passed |
| `tests/_agg/agg_http` | 1235 run, 1234 passed, 1 failed |
| `tests/_agg/agg_tls` | 117 run, 90 passed, 27 failed |
| `mojo format` | clean on all changed files |

Both aggregate failures are pre-existing here and reproduce identically on `d282574` with my branch stashed: `test_stream_coalescing` ("16 chunks should cost 1 writable edge, took 0") and 27 server-side cert-loading failures in `agg_tls`, the same 90/27 split on both. The `tests/certs` fixtures are valid to 2036, so it is not expiry; this is a local environment thing rather than anything either of us changed, and I did not chase it further. Everything Linux-side and CI-side is unverified from here.

ehsanmok:
Thanks! Sg. Please address the conflicts then will merge.

PR #15, opened 2026-08-28
feat(server): serve HTTP and WebSocket on one port

winding-lines:
`HttpServer` had no HTTP/1.1 upgrade path, and `WsServer` assumed every connection on its listener was a handshake. An app that wanted ordinary routes plus a WebSocket endpoint had to run two servers on two ports and put something in front of them.

`serve_ws_upgrade(handler, ws_handler)` makes it one listener. A request carrying `Connection: Upgrade` + `Upgrade: websocket` + a non-empty `Sec-WebSocket-Key` (RFC 6455 4.2.1) gets `101 Switching Protocols` and its socket goes to `ws_handler` as a `WsConnection`; everything else takes the unary `Handler.serve(req) -> Response` path untouched.

Not to be confused with `serve(handler, ws_handler)`, which is the WS-over-h2 sidecar (RFC 8441 CONNECT streams inside an h2 connection). Hence the separate name: this is the plain HTTP/1.1 handshake, and both can coexist.

The seam sits in `ConnHandle.on_readable`, inside the `headers.get("upgrade")` guard the h2c path already pays for, so non-upgrade traffic costs nothing new. It follows the h2c precedent throughout: `ws_handler` is an `Optional[WsHandlerFn]` on `ServerConfig`, a plain function pointer, so it copies into each per-worker config for free -- the property the standalone `WsServer` multi-worker path already relies on.

Two details worth calling out:

- The fd is detached from the `ConnHandle` (its `fd` set to -1) before the `WsConnection` takes it, so the handle's destructor cannot close a socket it no longer owns, and the reactor is told `done=True` so it unregisters rather than reaps.
- A client may pipeline its first frame into the same segment as the handshake. Those bytes are already in `read_buf` past the request and the reactor is about to stop reading that fd, so `WsConnection` gains a `prebuf` seeded from `read_buf[body_total:]` that `_recv_one` drains before its first socket read. Without it that frame is silently lost. The standalone `WsServer` path reads the handshake byte-at-a-time and leaves nothing buffered, so its prebuf is always empty.

`ws_handler` owns its connection for as long as it lives and blocks that worker meanwhile -- the model `WsServer` already uses. With one worker a long-lived socket therefore stalls the HTTP traffic behind it; the docstring says so and points at `num_workers > 1` or `WsServer`.

Cleartext only. `wss://` arrives through `TlsConnHandle`, which has no equivalent seam.

Tests: one server, one port -- an HTTP GET, then a WebSocket upgrade and echo round-trip, then another HTTP GET to show the port still serves HTTP afterwards. Plus the opt-in contract: a server started with plain `serve` must answer a handshake as an ordinary request and never send a
101. Example added under examples/intermediate/.

Created with Claude


flare/ws/server.mojo
```diff
@@ -414,6 +445,13 @@ struct WsConnection(Movable):
     def _recv_one(mut self) raises -> WsFrame:
         """Read bytes from stream and decode one complete frame."""
         var buf = List[UInt8](capacity=4096)
+        # Drain any bytes the HTTP reactor pre-buffered past the
+        # handshake (shared-listener upgrade path) before touching the
+        # socket. Empty for the standalone WsServer path.
+        if len(self._prebuf) > 0:
+            for i in range(len(self._prebuf)):
+                buf.append(self._prebuf[i])
+            self._prebuf.clear()
```
ehsanmok, line 534: `_prebuf` is a one-shot seed here, and `buf` is local, so anything `decode_one` didn't consume goes away with `buf` when this returns. `_DecodeResult.consumed` at `flare/ws/frame.mojo:94` is right there and nothing reads it.

Concretely: a client that pipelines the handshake plus two frames into one segment gets the first frame delivered and the second dropped, and the next `recv()` then blocks on a socket with nothing left to send. `recv`'s PING branch at line 441 hits the same thing with a PING+TEXT pair, since it consumes the PING and loops.

Making `_prebuf` the persistent carry-over fixes both: take it into `buf` here as you already do, then put `buf[result.consumed:]` back into it before returning the frame. Costs nothing on the standalone `WsServer` path, where it stays empty throughout.

flare/ws/server.mojo
```diff
@@ -337,10 +337,41 @@ struct WsConnection(Movable):
 
     var _stream: TcpStream
     var _peer: SocketAddr
+    var _prebuf: List[UInt8]
+    """Bytes already drained from the socket before this
+    ``WsConnection`` took ownership of the fd. Non-empty only on the
+    shared-listener upgrade path (``HttpServer.serve(handler,
+    ws_handler)``), where the HTTP/1.1 reactor may have buffered
```
ehsanmok: `HttpServer.serve(handler, ws_handler)` is the RFC 8441 h2 sidecar, which is the confusion this PR spends three paragraphs heading off. The overload that fills this buffer is `serve_ws_upgrade`.

```suggestion
    shared-listener upgrade path (``HttpServer.serve_ws_upgrade``),
    where the HTTP/1.1 reactor may have buffered
```

flare/http/_reactor/conn_handle.mojo
```diff
@@ -1180,6 +1205,87 @@ struct ConnHandle(Movable):
...
+        unregisters the fd; closing it is the ``WsConnection``'s job.
+
+        Cleartext only. A ``wss://`` connection arrives through
+        ``TlsConnHandle``, which has no equivalent seam, so a
+        TLS-terminated server still needs a separate ``WsServer``.
+        """
+        from flare.http.server import _ascii_lower
+
+        # RFC 6455 4.2.1 qualification.
+        var upg = _ascii_lower(req.headers.get("upgrade"))
```
ehsanmok, line 1248: `WsServer`'s reader skips the request line, so it has no method to check. Here the parsed `Request` is in hand, and as written a `POST /ws` carrying the three headers gets a 101. RFC 6455 4.1 wants GET on HTTP/1.1 or later. `== "HTTP/1.0"` is the tree's idiom for that version test, see `flare/http/_reactor/keepalive_scan.mojo:374`.

```suggestion
        # RFC 6455 4.2.1 qualification.
        if req.method != "GET" or req.version == "HTTP/1.0":
            return False
        var upg = _ascii_lower(req.headers.get("upgrade"))
```

Optional in this PR. If you'd rather keep it to the seam, I'll fold this into the `Sec-WebSocket-Version` follow-up along with the same gap in `WsServer`.

tests/http/test_server_ws_upgrade.mojo
```diff
@@ -0,0 +1,254 @@
...
+    )
+    assert_true(
+        "hello http" in body,
+        "upgrade request did not reach the HTTP handler; got: " + body,
+    )
+
+
+def main() raises:
+    test_http_and_ws_on_one_port()
+    test_upgrade_request_is_ordinary_traffic_without_a_ws_handler()
```
ehsanmok, line 382: `WsClient.connect` writes the handshake, waits for the 101, and only then writes the frame in a separate `send`, so on loopback those land as two segments and `read_buf[body_total:]` is empty. Both tests here run with an empty `_prebuf`, which means the new constructor, the slice at `flare/http/_reactor/conn_handle.mojo:1269` and the drain at `flare/ws/server.mojo:451` are exercised only as no-ops.

A third test that pushes the handshake and a masked TEXT frame through `_connect_loopback` in one `_send`, then reads the 101 and the echo off the same fd, covers it. Put two frames in that write and it also pins down the leftover-bytes problem I raised on `_recv_one`.

flare/http/_reactor/conn_handle.mojo
```diff
@@ -634,6 +634,31 @@ struct ConnHandle(Movable):
...
+                    self.should_close = True
+                    return StepResult(
+                        want_read=False, want_write=False, done=True
+                    )
+                if upgraded:
+                    return StepResult(
+                        want_read=False,
+                        want_write=False,
+                        done=True,
+                        idle_timeout_ms=0,
```
ehsanmok, line 660: Dead as written. Both reactors only reach `_apply_step` when `not step_done` (`flare/http/_unified_reactor_impl.mojo:225`, `flare/http/_server_reactor_epoll.mojo:445`), and this result has `done=True`, so the timer for this fd is dropped by the cleanup path a moment later regardless. Drop the argument so the line stops reading as if it cancels something.

ehsanmok (changes requested):
Thanks for this - one listener serving both protocols closes a real gap, and the shape is right. The seam sits inside the `upgrade` guard the h2c path already pays for, the fd detach copies `_migrate_h1_to_h2` almost line for line, and the handshake reuses `_compute_accept_srv` / `_send_upgrade_response` instead of growing a second copy. I built the branch and ran everything you listed; it all passes. Two things stop me merging as-is, both about the prebuf, which is the one piece of genuinely new wire behaviour in the diff: it only survives a single frame, and nothing in the suite exercises it at all. Everything else below I'll take myself.

On the rebase: this branch predates the 2026-09-11 migration on `main` and conflicts only in `pixi.toml`, where the resolution isn't textual because the `tests` chain you appended to no longer exists (`tests = "bash tools/run_test_aggregates.sh"`, `pixi.toml:653`). Run `pixi run tests-gen` after rebasing so `tests/http/test_server_ws_upgrade.mojo` joins `tests/_agg`; see my note on #11 for what that runner does. `flare/http/_reactor/conn_handle.mojo`, `flare/http/server.mojo` and `flare/ws/server.mojo` auto-merge, though the 843-site API migration means they may still want a compile pass afterwards.

## What I verified

Worktree at 5b96dd5, Mojo 1.0.0 (ed45d567), macOS arm64.

| entry | result |
|---|---|
| `tests/http/test_server_ws_upgrade.mojo` | 2 passed |
| `tests/http/test_server.mojo` | 84 run, 84 passed, 0 failed |
| `tests/ws/test_ws.mojo` | 34 run, 34 passed, 0 failed |
| `tests/http/test_h2c_upgrade.mojo` | 7 passed |
| `tests/ws/test_ws_h2_reactor.mojo` | 1 passed |
| `examples/intermediate/http_ws_one_port.mojo` | compiles |

`mojo format` clean on all six changed files.

Things I chased down and found correct, so nobody has to re-argue them:

- Ordering against h2c. `Upgrade: h2c` fails `upg != "websocket"` at `flare/http/_reactor/conn_handle.mojo:1245` and falls through to `_h2c_upgrade_decode_settings` untouched; `Upgrade: websocket` carrying an `HTTP2-Settings` header takes the WS path, which is what it should do. `test_h2c_upgrade` still green.
- The `read_buf[body_total:]` slice. `body_total` is `headers_end + content_length` (`conn_handle.mojo:479`) or the chunked end (`:501`), and `_finalise_response` compacts the prefix away after every request (`:544`), so by the time the upgrade request is dispatched `read_buf[body_total:]` is exactly the bytes that follow it. For a GET with no body that is `headers_end`. An earlier pipelined request cannot straddle the boundary.
- The 101 write. `raw.set_nonblocking(False)` at `conn_handle.mojo:1281` runs before `_send_upgrade_response`, so `stream.write_all` is on a blocking fd and there is no EAGAIN partial-write window. Worth stating out loud, because the ordering of those two lines is load-bearing and nothing enforces it.
- The fd handoff. `_cleanup_conn_unified` (`flare/http/_unified_reactor_impl.mojo:341`) unregisters by the dispatch `fd` taken from the event token, not by `_stream._socket.fd`, so zeroing the field doesn't break the unregister. The `WsConnection` destructor does close the fd before that unregister runs, so `reactor.unregister` sees a closed fd - harmless here, because the worker is single-threaded (nothing can claim that fd number in between), both epoll and kqueue drop a closed fd on their own, and the call already sits in a `try`. I went looking for what #16 does with that window, since it moves the close onto another thread, and wrote up what I found there.
- The function-pointer claim. `ServerConfig` derives `Copyable` with no hand-written `copy`, `_serve_multicore` passes `self.config.copy()` (`flare/http/server.mojo:1230`), and `serve_ws_upgrade` sets the field before calling `serve`, so `num_workers > 1` with SO_REUSEPORT carries `ws_handler` into every worker. Read, not run.
- The qualification checks match the standalone parser exactly. `_parse_ws_upgrade_bytes` and `_read_upgrade_request` (`flare/ws/server.mojo:162`, `:235`) test the same three things: `Upgrade` equal to `websocket` after lowercasing, `upgrade` as a substring of `Connection`, non-empty `Sec-WebSocket-Key`. The two paths accept and reject the same requests, which is the property I cared about most going in.

Could not verify: the multi-worker upgrade (no test, and I didn't write one), the `use_bufring` interaction in item 4, and anything TLS.

## 1. The prebuf only carries the first frame

This is the blocking one. `_recv_one` builds a fresh local `buf` on each call, drains `_prebuf` into it, decodes one frame, and then throws `buf` away (`flare/ws/server.mojo:445-457`). `_DecodeResult` carries a `consumed` count (`flare/ws/frame.mojo:94`) that nothing reads, so whatever sits in `buf` past the returned frame is dropped.

The guarantee the commit message states - that a pipelined first frame is never lost - therefore holds for exactly one frame. A client that puts the handshake plus two frames in one segment gets the first delivered and the second dropped; a handshake plus PING plus TEXT does the same, since `recv`'s PING branch at `:441` consumes the PING and loops. The next `recv()` then blocks on a socket with nothing more coming, so the failure mode is a hang, with no error.

It's pre-existing in `WsServer`, where it was hard to reach: the byte-at-a-time handshake reader leaves nothing buffered, so every `_recv_one` starts empty. This PR is what makes it reachable with real data, so I'd rather it went in here than trailed behind.

The fix is small and `_prebuf` is already the right home for it - make it the connection's persistent carry-over buffer rather than a one-shot seed. Take it into `buf` at the top of `_recv_one` as you do now, and on a successful decode put `buf[result.consumed:]` back into `_prebuf` before returning the frame. Then `recv`'s PING loop and every later `_recv_one` pick up where the previous one stopped, and the standalone path is unaffected because the buffer stays empty there. I'm not posting that as a suggestion block because I'm unsure how cleanly `^`-moving the field out and back reads in Mojo here, and I'd rather you pick the spelling.

## 2. Nothing exercises a non-empty prebuf

Same subject, separate ask. Both tests drive the WebSocket through `WsClient.connect`, which writes the handshake, reads the 101, and only then writes the frame in a second `send`. On loopback those are two segments, so `read_buf[body_total:]` is empty and `_prebuf` is empty in both tests. The new constructor at `flare/ws/server.mojo:358`, the slice at `conn_handle.mojo:1269-1272` and the drain at `flare/ws/server.mojo:448-454` are all untested - they happen to run as no-ops.

A third test that writes the handshake and a masked TEXT frame in one `send()` on a raw socket, then reads the 101 and the echo off the same fd, covers it; the `_connect_loopback` helper is already in the file. Send two frames in that one write and it pins down item 1 as well. New wire-level behaviour wants a test before merge, and this is the one bit of the PR that has none.

## 3. The prebuf docstring names the wrong entry point

`flare/ws/server.mojo:343` says the prebuf is non-empty on `HttpServer.serve(handler, ws_handler)`. That overload is the RFC 8441 h2 sidecar, which is the confusion the rest of this PR spends three paragraphs heading off. The path that fills the prebuf is `serve_ws_upgrade`. Suggestion inline.

## 4. Follow-ups I'll take, not asks

- `Sec-WebSocket-Version` is never checked, so a client offering version 8 gets a 101 and a protocol mismatch instead of the 426 plus `Sec-WebSocket-Version: 13` that RFC 6455 4.4 asks for. `WsServer` doesn't check it either, so the tree is at least consistent, and I'd rather fix both at once than fork the behaviour here.
- Method and HTTP version are unchecked too. Unlike `WsServer`, which skips the request line entirely, you have the parsed `Request` in hand, so `POST /ws` with the three right headers gets a 101. That one is four lines and there's a suggestion inline; take it or leave it, I'll pick it up either way.
- The seam lives only in `on_readable`. `on_readable_from_buf`, `on_readable_static`, `on_readable_view` and `on_readable_cancel` don't have it, so a server with `config.use_bufring` on Linux silently answers handshakes as ordinary HTTP. h2c has the identical hole today, so I'm not treating it as a regression, but the `serve_ws_upgrade` docstring should say which reader paths it covers.
- TLS. `TlsConnHandle` has no equivalent seam and a `wss://` server still needs a separate `WsServer`. The docstring says so, which is all I want for now.
- `idle_timeout_ms=0` on the `done=True` result at `conn_handle.mojo:660` is never read. Noted inline; harmless, just misleading.

The example is honest about what it demonstrates and matches the API. The commit message describes the code accurately apart from the prebuf guarantee in item 1, which overstates by one frame - worth a word there too when you respin.

## What I'll take

The four items in section 4: the `Sec-WebSocket-Version` check across both parsers, the method and version qualification if you'd rather not take the suggestion inline, the reader-path coverage wording on the `serve_ws_upgrade` docstring, and the dead `idle_timeout_ms` argument.

From you: items 1 and 2, the docstring in item 3, and the rebase.

winding-lines:
Thanks — items 1 and 2 were the same defect seen from two sides, and the test turned the hang into a failure. Rebased onto `d282574`, one commit, `pixi run tests-gen` run.

The `pixi.toml` conflict resolved to main's side; the `test-server-ws-upgrade` convenience task moved next to `test-ws-h2-reactor` to match the ~100 others still there. Say the word if you would rather it went the way you asked for `test-server-ws-offload` on #16 and I will drop it from both.

`flare/ws/server.mojo` also conflicted, not just auto-merged — #11 landed `origin` in the same two places this adds `_prebuf`. Both fields and both constructors coexist now.

## 1. `_prebuf` is the carry-over, not a seed

Took your shape. `_recv_one` drains it into `buf` as before, and on a successful decode puts `buf[result.consumed:]` back before returning the frame. `_DecodeResult.consumed` finally has a reader.

The spelling, since you left it to me: a plain `for i in range(result.consumed, len(buf)): self._prebuf.append(buf[i])` just before `return result^.take_frame()`. `_prebuf` is empty at that point — drained at the top, and the `except` branch only ever appends to `buf` — so there is nothing to move out and back, and no `^` dance to get wrong. `take_frame` consumes `result`, so `consumed` has to be read first, which the ordering handles on its own.

`recv()`'s PING loop picks up the carry-over on its next `_recv_one` for free, and the standalone `WsServer` path is untouched because the buffer stays empty there throughout.

## 2. A test that reaches it

`test_frames_pipelined_with_the_handshake_are_all_delivered`, through `_connect_loopback` as you suggested: handshake plus two masked TEXT frames in one `send`, then the 101 and both echoes off the same fd. Two frames, so it pins item 1 and not just the drain.

It does not hang. Without the put-back it deadlocks exactly as you described — the server sits in `recv()` on a frame it already dropped while the client waits for an echo that will never come — so the socket carries a 3 s `SO_RCVTIMEO` and the pre-fix result is:

```
AssertionError: second pipelined frame was dropped; got: HTTP/1.1 101 Switching Protocols ...
```

I verified that by reverting the two-line put-back and re-running: hang before the timeout was added, clean failure after.

## 3

Docstring now says `HttpServer.serve_ws_upgrade`, and describes the field as the persistent carry-over rather than a one-shot seed.

## Section 4

Took the method and version suggestion inline — `if req.method != "GET" or req.version == "HTTP/1.0"` ahead of the header checks, using the `keepalive_scan.mojo:374` idiom. The rest are yours as offered: `Sec-WebSocket-Version` across both parsers, the reader-path coverage wording, and the dead `idle_timeout_ms`.

One thing you may want in the `Sec-WebSocket-Version` follow-up while you are in there: `WsConnection.origin` is empty on this path. Both `_prebuf` constructors set it, and the seam has the parsed `Request` in hand, so `req.headers.get("origin")` is right there — but populating it is the same call you already made for `WsServer`, and I would rather you made it once for both than have me make it differently here. Flagging it because after #11 a handler that checks `conn.origin` reads `""` for every connection served this way.

## Commit message

Rewritten. The prebuf paragraph said the guarantee held for a pipelined first frame; it now says what actually holds and what used to happen instead.

## What I ran

macOS arm64, Mojo 1.0.0 (ed45d567), at `07b3060`.

| entry | result |
|---|---|
| `tests/http/test_server_ws_upgrade.mojo` | 3 passed |
| `tests/ws/test_ws.mojo` | 38 run, 38 passed |
| `tests/http/test_h2c_upgrade.mojo` | 7 passed |
| `tests/ws/test_ws_h2_reactor.mojo` | 1 passed |
| `tests/ws/test_ws_multicore.mojo` | 1 passed |
| `tests/http/test_server_multicore.mojo` | 4 run, 4 passed |
| `tests/_agg/agg_ws` | 78 run, 78 passed |
| `tests/_agg/agg_http` | 1232 run, 1231 passed, 1 failed |
| `examples/intermediate/http_ws_one_port.mojo` | compiles |
| `mojo format` | clean on all changed files |

The one `agg_http` failure is `test_stream_coalescing`, pre-existing: it reproduces identically on `d282574` with this branch stashed. Compiles clean after the rebase onto the API migration. Nothing verified on Linux.

PR #16, opened 2026-08-28
feat(server): opt-in ws_offload, one thread per WebSocket

winding-lines:
**Stacked on `feat/ws-upgrade-one-port` — review that one first.**

`_handle_ws_upgrade` calls `ws_handler(conn)` on the reactor thread, so the worker is parked for the WebSocket's entire lifetime. That is the model `WsServer` already uses and it is fine when handlers are short, but a long-lived connection head-of-line-blocks every other fd pinned to that worker — ordinary keep-alive HTTP requests included. On macOS the `SO_REUSEPORT` hash cheerfully lands them on the busy worker.

`ServerConfig.ws_offload` (also a `serve_ws_upgrade` argument) hands the already-detached, blocking-mode fd to a fresh detached thread via `spawn_ws_offload` and returns, so the reactor keeps serving other connections **even with a single worker**. Default `False`, so the existing path is byte-identical.

### Why no cross-thread write marshaling is needed

The connection stays on that one thread from handshake to close — every `recv` / `send_text` / `close` and the final socket close in `WsConnection.__del__`. So `WsConnection`'s non-thread-safe writes are never touched concurrently, and there is no outbox to build.

The costs are stated in the docstring rather than buried: one thread per concurrent WebSocket **with no built-in cap**, and handlers that used to be serialised per worker now run concurrently, so shared state is the handler's problem.

`ThreadHandle.detach()` wraps `pthread_detach` and zeroes `_thread_id` so a later `join()` short-circuits rather than joining a detached thread (undefined per POSIX). Without detaching, a never-joined joinable thread holds its kernel bookkeeping until the process exits — one leak per connection.

### Tests

`tests/http/test_server_ws_offload.mojo` — a differential on the same server and the same deliberately slow handler. A plain HTTP GET issued while a WebSocket is mid-handler:

| `ws_offload` | GET round-trip |
|---|---|
| `True` | **0–1 ms** |
| `False` | **~2300 ms** |

against a 2500 ms hold and a 1200 ms split — stable across repeated runs, so the margins are nowhere near flaky. The inline case is asserted too, so the blocking this fixes cannot quietly regress into looking like the fixed case.

Verified against this branch's tree with Mojo 1.0.0; `test_server_ws_upgrade`, `test_ws` (34), `test_ws_multicore`, `test_thread_ffi` (10) and `test_server` (84) all still pass.


flare/ws/server.mojo
```diff
@@ -677,6 +715,83 @@ def _handle_ws_connection(
...
+    debug_assert[assert_mode="safe"](
+        Int(ctx_ptr) != 0,
+        "spawn_ws_offload: alloc[_WsOffloadCtx] returned NULL",
+    )
+    ctx_ptr.init_pointee_move(_WsOffloadCtx(conn^, handler))
+    var arg = UnsafePointer[UInt8, MutUntrackedOrigin](
+        unsafe_from_address=Int(ctx_ptr)
+    )
+    var th = ThreadHandle.spawn[_ws_offload_entry](arg)
+    th.detach()
```
ehsanmok: If `pthread_create` fails here, `_ws_offload_entry` never runs, so nothing takes the context back out of the heap slot and nothing destroys the `WsConnection` inside it. The fd stays open with no owner - the reactor already set its own copy to `-1` at `flare/http/_reactor/conn_handle.mojo:1279` - and the error surfaces at `conn_handle.mojo:647` as a generic handshake failure.

`EAGAIN` from a process at its thread limit is the reachable case, and with no cap on offloaded threads a client can drive the server there and then leak one live socket per handshake until it hits `EMFILE`.

`unsafe_deinit_pointee()` then `free()` is the pair `_ws_serve_multicore` uses at `flare/ws/server.mojo:929`.

```suggestion
    var th: ThreadHandle
    try:
        th = ThreadHandle.spawn[_ws_offload_entry](arg)
    except e:
        ctx_ptr.unsafe_deinit_pointee()
        ctx_ptr.free()
        raise e
    th.detach()
```

flare/http/_reactor/conn_handle.mojo
```diff
@@ -1180,6 +1205,96 @@ struct ConnHandle(Movable):
...
+        2. **Detached** the fd from ``self._stream`` -- its ``fd`` is
+           set to ``-1`` so this handle's destructor will not close a
+           socket it no longer owns -- and handed it to a
+           :class:`flare.ws.WsConnection`, seeded with any frame bytes
+           the reactor had already read past the handshake request.
+        3. Put the socket back into blocking mode, because the WS
+           handler uses blocking ``recv`` exactly as
+           :class:`flare.ws.WsServer` does.
+        4. Run ``ws_handler(conn)``, which owns that connection for as
+           long as it lives.
```
ehsanmok: This still describes the inline path only. With `config.ws_offload` set, the method returns as soon as the thread is spawned - the handler is still running - so `True` no longer means the handler finished. Item 4 has the same problem.

```suggestion
        Returns ``True`` once the handshake has been answered. By then
        this method has:

        1. Computed ``Sec-WebSocket-Accept`` and written ``101
           Switching Protocols`` straight to the socket.
        2. **Detached** the fd from ``self._stream`` -- its ``fd`` is
           set to ``-1`` so this handle's destructor will not close a
           socket it no longer owns -- and handed it to a
           :class:`flare.ws.WsConnection`, seeded with any frame bytes
           the reactor had already read past the handshake request.
        3. Put the socket back into blocking mode, because the WS
           handler uses blocking ``recv`` exactly as
           :class:`flare.ws.WsServer` does.
        4. With ``config.ws_offload`` unset, run ``ws_handler(conn)``
           to completion on this thread. With it set, moved the
           connection onto a detached thread and returned while the
           handler is still running.
```

flare/http/server.mojo
```diff
@@ -750,6 +751,81 @@ struct HttpServer(Movable):
...
+        ``ws_handler`` has the same signature as
+        :meth:`flare.ws.WsServer.serve`'s callback, so a handler
+        written against a standalone ``WsServer`` moves over unchanged.
+        It owns its connection for as long as that connection lives and
+        blocks the reactor worker for that whole time -- the model
+        ``WsServer`` already uses. With one worker, one long-lived
+        WebSocket therefore stalls the HTTP traffic behind it; give the
+        server ``num_workers > 1``, or reach for
+        :meth:`flare.ws.WsServer` when connections are many and
+        long-lived.
```
ehsanmok: The `ws_offload` entry you added under `Args:` is accurate. This paragraph above it still tells readers the handler blocks the reactor worker for the connection's whole lifetime and sends them to `num_workers > 1` or a standalone `WsServer`. Your own PR description argues `num_workers > 1` does not help - the `SO_REUSEPORT` hash lands new connections on the busy worker as happily as anywhere else.

```suggestion
        It owns its connection for as long as that connection lives.
        By default it runs inline on the reactor worker and blocks it
        for that whole time, the model ``WsServer`` already uses, so
        with one worker a long-lived WebSocket stalls the HTTP traffic
        behind it. Pass ``ws_offload=True`` to move each upgraded
        connection onto its own thread, or reach for
        :meth:`flare.ws.WsServer` when connections are many and
        long-lived.
```

flare/http/_server/config.mojo
```diff
@@ -126,6 +139,53 @@ struct ServerConfig(Copyable, Movable):
...
+    ``True`` hands the (already-detached, blocking-mode) fd to a fresh
+    detached thread via :func:`flare.ws.server.spawn_ws_offload` and
+    returns, so the worker keeps serving other connections while the
+    WebSocket runs to completion off-reactor. The connection stays on
+    that one thread start to finish, so ``WsConnection``'s
+    non-thread-safe writes are never touched concurrently.
+
+    The costs: one thread per concurrent WebSocket, with no built-in
+    cap, and handlers that used to be serialised per worker now run
+    concurrently -- any state they share is theirs to protect.
```
ehsanmok, line 185: Worth being specific about what "no built-in cap" means for someone who has already set `max_connections`. That cap bounds `len(conns)` in the accept loop (`flare/http/_server_reactor_epoll.mojo:247`, `flare/http/_unified_reactor_impl.mojo:662`), and an upgraded connection leaves that table on the same step it is handed to the thread, so it stops counting. `max_connections=1000` with `ws_offload` on still allows an unbounded number of live WebSockets.

```suggestion
    The costs: one thread per concurrent WebSocket, with no built-in
    cap, and handlers that used to be serialised per worker now run
    concurrently -- any state they share is theirs to protect.

    An offloaded connection leaves the reactor's live-connection table
    as soon as the handshake completes, so
    :attr:`ServerConfig.max_connections` does not bound it. Until a cap
    exists, leave this off on a listener reachable by untrusted peers:
    N concurrent handshakes means N pthreads, each with a
    platform-default stack.
```

flare/runtime/_thread.mojo
```diff
@@ -174,6 +174,32 @@ struct ThreadHandle(Movable):
         # further pthread_join on the stale id is UB per POSIX.
         self._thread_id = UInt64(0)
 
+    def detach(mut self) raises:
```
ehsanmok, line 181: `detach()` makes "a thread that outlives its handle" the supported case, which contradicts the module header at `flare/runtime/_thread.mojo:20-21` ("Threads that outlive their `ThreadHandle` are undefined behaviour. Always join before dropping.") and `spawn`'s `Returns:` at :117 ("A `ThreadHandle` the caller must `join()`"). Both are outside the diff, so I cannot suggest on them. Wording I'd use for the header:

```
Threads that outlive their ``ThreadHandle`` are undefined behaviour
unless the handle was ``detach()``ed first. Join before dropping, or
detach and let the OS reclaim the thread.
```

The rest of the contract holds, and I checked each part: `join()` short-circuits on a zeroed id at :160, `detach()` does the same at :196 so a second detach is a no-op, and `ThreadHandle` has no destructor, so dropping a detached handle does nothing at all.

ehsanmok (changes requested):
Thanks for this - the head-of-line blocking is real, and the differential test is the right shape for proving it. I pulled the branch and reproduced everything before reviewing. One thing has to change before I merge: `spawn_ws_offload` drops the connection on the floor, fd and all, when `pthread_create` fails, and that is exactly the failure the missing cap invites. The rest is docstrings that still describe the old behaviour, plus two items I'm happy to take myself.

On the rebase: this conflicts with `main` only in `pixi.toml`, same as #15 under it. Run `pixi run tests-gen` after rebasing so `tests/http/test_server_ws_offload.mojo` joins `tests/_agg` and the hand-written `test-server-ws-offload` task and `tests` chain entry can both go; see my note on #11 for what that runner does. `flare/http/_reactor/conn_handle.mojo`, `flare/http/server.mojo`, `flare/runtime/_thread.mojo` and `flare/ws/server.mojo` auto-merge cleanly, but the deprecated-API migration landed after this branch, so I want to see it compile once rebased.

## What I ran

Mojo 1.0.0 (ed45d567), macOS arm64, against the branch head 3e00a27.

| entry | result | wall |
|---|---|---|
| `tests/http/test_server_ws_offload.mojo` | 2 passed | 17 s |
| `tests/http/test_server_ws_upgrade.mojo` | 2 passed | 12 s |
| `tests/runtime/test_thread_ffi.mojo` | 10 run, 10 passed, 0 failed | 2 s |
| `tests/ws/test_ws_multicore.mojo` | 1 passed | 4 s |
| `tests/http/test_server_multicore.mojo` | 4 run, 4 passed, 0 failed | 4 s |

`mojo format` is clean on all 8 changed `.mojo` files.

What I could not verify: anything Linux-side - this box is macOS arm64, so the glibc default stack size and `pin_to_cpu` are read rather than measured; the `pthread_create` failure path in item 1, because I have no way to drive the process to its thread limit from a test; CI timing on a loaded GitHub runner; and whether the branch still compiles after the rebase onto `main`'s API migration.

One thing I went looking for and found fine, because I flagged it on #15. There the `WsConnection` destructor closes the fd before `_cleanup_conn_unified` (`flare/http/_unified_reactor_impl.mojo:341`) unregisters it, and I called that window harmless only because the worker is single-threaded. Offloading reverses the order rather than widening it: `_handle_ws_upgrade` returns `done=True` as soon as the thread is spawned, so the reactor unregisters at `_unified_reactor_impl.mojo:356` while the fd is still open on the offload thread, and the thread closes it later with nothing watching. The fd number cannot be recycled under the unregister while the socket is still open, so the hazard I left open on #15 does not follow the connection onto the thread. Item 1 is a different problem: there the socket is never closed at all.

## 1. `spawn_ws_offload` leaks the connection when `pthread_create` fails

At `flare/ws/server.mojo:782` the context is heap-allocated and the `WsConnection` is moved into it. `ThreadHandle.spawn` at :791 raises when `pthread_create` returns non-zero, and on that path `_ws_offload_entry` never runs, so nothing takes the `_WsOffloadCtx` back out of the slot and nothing frees it. The `WsConnection` inside is never destroyed, so its `__deinit__` never runs and the socket stays open. The reactor already zeroed its own copy of the fd at `flare/http/_reactor/conn_handle.mojo:1279`, so nobody else closes it either. The error lands in the `except` at `conn_handle.mojo:647`, which sets `should_close` on a handle whose fd is already `-1`.

The trigger is the case this PR already says it does not guard. `pthread_create` returns `EAGAIN` once the process is at its thread limit, so the first N WebSockets consume the threads and every handshake after that leaks a live socket permanently. Those fds are not reclaimed when the peer goes away - the fd is still open and nothing polls it. A client opening connections in a loop takes the server to `EMFILE` and leaves it there. Suggestion inline - `unsafe_deinit_pointee()` then `free()` is the pair `_ws_serve_multicore` already uses at `flare/ws/server.mojo:929`.

One second-order note I'd take as a comment rather than code: if `th.detach()` at :792 raises, the thread is already running and owns the context, so re-raising is correct there and a free would be a double free. Worth remembering before anyone extends the handling.

## 2. Three docstrings now describe the inline path only

This is the other reason I'm holding rather than merging with a note - docstrings here have to say what the code does now.

- `flare/http/_reactor/conn_handle.mojo:1217` says `True` means "the handshake has been answered and the user's `ws_handler` has run to completion", and item 4 at :1231 says the same. Under `ws_offload` the method returns while the handler is still running. Suggestion inline.
- `flare/http/server.mojo:776-782`. `serve_ws_upgrade`'s prose still says the handler blocks the reactor worker for the connection's whole lifetime and points readers at `num_workers > 1` or a standalone `WsServer`. I left that paragraph alone on #15 because it was accurate there; this PR is what makes it wrong. That paragraph describes the problem this PR fixes, and by your own argument in the description `num_workers > 1` is not the fix - the `SO_REUSEPORT` hash lands new connections wherever it likes. Suggestion inline.
- `flare/runtime/_thread.mojo:20-21`. The module header still reads "Threads that outlive their `ThreadHandle` are undefined behaviour. Always join before dropping." `detach()` makes outliving the handle the supported case, and `spawn`'s `Returns:` at :117 repeats the old rule. Both sit outside the diff, so the wording I'd use is in the inline comment on `detach()` instead.

The caller comment at `conn_handle.mojo:640-641` ("On success the handler has run to completion") has the same problem. That line came in with #15, so fold it in wherever it is easiest.

## 3. The cap, and what `max_connections` does not cover

I'm fine deferring the cap itself. I do want the docstring to say more than "no built-in cap", because there is a specific thing a reader will assume and be wrong about. `ServerConfig.max_connections` bounds `len(conns)` in the accept loop (`flare/http/_server_reactor_epoll.mojo:247`, `flare/http/_unified_reactor_impl.mojo:662`). An upgraded connection returns `done=True` and leaves that table on the same step it is handed to the thread, so it stops counting against the cap the moment the handshake completes. `max_connections=1000` with `ws_offload` on still allows an unbounded number of live WebSockets - a thousand in the reactor table at a time, and whatever has accumulated behind it.

The threads are also default-sized. `ThreadHandle.spawn` passes a NULL attr at `flare/runtime/_thread.mojo:131`, so each offloaded connection gets the platform default stack, which is 512 KiB on macOS and `RLIMIT_STACK` on glibc, commonly 8 MiB. That is reserved address space, so the thread count and the resulting `EAGAIN` bite first - which loops straight back to item 1.

What I want here is the docstring warning (suggestion inline) and a follow-up issue covering the cap and `pthread_attr_setstacksize`. I'll open the issue if you'd rather not.

## 4. Shutdown, which I'll take myself

`drain(timeout_ms)` at `flare/http/server.mojo:1602` closes the listener, waits for in-flight reactor events to flush and breaks the loop. An offloaded thread is not a reactor event, so drain returns without waiting for it and the `ShutdownReport` counts it as nothing. Process exit then kills those threads mid-frame with no close frame sent. Inline WebSockets have the same gap today - the handler runs to completion and drain cannot interrupt it - so this is not a regression. Detaching does make it harder to fix later, because there is nothing left to join.

Nothing needed from you in this PR. I'll open the issue alongside the cap one.

## 5. Tests

The differential is well built, and I checked the thing I most wanted to check: the inline case cannot pass by accident. `test_inline_websocket_blocks_http_on_one_worker` asserts `"hello http" in body` before it asserts anything about timing, and `_measure` returns `-1` when anything inside its `try` throws, which fails `elapsed >= _SPLIT_MS`. A GET that fails instead of blocking fails the test.

Margins look right to me. The handler holds for 2500 ms, the split is at 1200 ms and the GET goes out roughly 200 ms in, so the inline case has about 1100 ms of slack and the offload case about 1200 ms. Wide enough for a slow runner. The file took 17 s of wall here - nearly all of it compile.

Two gaps, neither of them blocking:

- Nothing asserts that two offloaded WebSockets run at the same time. The current test proves the reactor stays free. It says nothing about two handlers overlapping, and concurrency is the part of the docstring that changes what handler authors have to do.
- Nothing covers the failure path in item 1. Forcing `EAGAIN` from a test is awkward, so I'm not asking for it.

## Minor

`spawn_ws_offload` is the only non-underscore name in that neighbourhood - `_compute_accept_srv`, `_send_upgrade_response` and `_ws_offload_entry` around it are all private, and it is not re-exported from `flare/ws/__init__.mojo`. `_spawn_ws_offload` would match, unless you meant it as public surface.

`_ws_offload_entry` builds its NULL return inline at `flare/ws/server.mojo:766`. `_null_ptr()` at `flare/runtime/_thread.mojo:63` exists for exactly that and carries the explanation of why the address has to come from a runtime zero.

## What I'll take

The cap and stack-size issue, the shutdown issue, and the `_thread.mojo` module header if you'd rather not touch a file outside the feature.

From you: the free on the `pthread_create` failure path (item 1), the three docstrings (item 2), the `max_connections` paragraph in the `ws_offload` docstring (item 3), and the rebase.

The commit message reads accurate against the diff - and the `ThreadHandle.detach()` rationale in it is exactly right, and exactly where it belongs.

winding-lines:
Thanks — item 1 was a real leak and the trigger is exactly the one the missing cap invites. Rebased onto the respun #15, which is itself on `d282574`; it compiles after the API migration. `pixi run tests-gen` run, and the hand-written `test-server-ws-offload` task and `tests` chain entry are both gone as you asked.

## 1. The free on the `pthread_create` failure path

Took your suggestion. `ThreadHandle.spawn` is wrapped, and the `except` does `unsafe_deinit_pointee()` then `free()` before re-raising — the pair `_ws_serve_multicore` uses at :929. Destroying the context runs `WsConnection.__deinit__`, which closes the fd, so the socket the reactor already disowned goes with it instead of accumulating one per handshake until `EMFILE`.

Your second-order note is in the code as a comment rather than as handling, since it is a thing to remember and not a thing to do: no free on the `detach()` path, because by then the thread is running and owns the context and a free there would be a double free once `_ws_offload_entry` takes the pointee.

The docstring's `Raises:` says why the free is there, so nobody deletes it as redundant.

## 2. The three docstrings

- `conn_handle.mojo:1217` — took your suggested block verbatim, including the reworked item 4.
- `flare/http/server.mojo` — took the suggested paragraph. It no longer sends readers to `num_workers > 1`.
- `flare/runtime/_thread.mojo` — took your header wording. `spawn`'s `Returns:` now says "join() or detach()", and its `Raises:` says nothing was spawned so `arg` is still the caller's to free, which is the contract item 1 depends on.

The caller comment at `conn_handle.mojo:640` is folded in too.

## 3. The `max_connections` paragraph

Took your suggestion verbatim.

Deferring the cap itself and the stack size to you, with the issue — thanks for offering to open it.

## Minor

`spawn_ws_offload` is now `_spawn_ws_offload`; it was never meant as public surface. `_ws_offload_entry` uses `_null_ptr()`.

## 5. The concurrency gap

Took that one. `test_two_offloaded_websockets_run_concurrently`: two connections, each handler holding 2500 ms. Measured 2510 ms against 5000 ms serialised, with the split at 4000 ms — about 1500 ms of slack either way, same discipline as `_SPLIT_MS`.

It is worth having precisely because the other two cannot fail it: a single shared handler thread would satisfy "the reactor stays free" and still serialise every WebSocket. And it is the assertion behind the docstring line telling handler authors their shared state is now their problem.

I did not write the `EAGAIN` case. Forcing it from a test is as awkward as you said.

One honest caveat on that test: with `ws_offload=False` it does not fail, it deadlocks — the second `WsClient.connect` never completes, because the first handler owns the worker from the moment of upgrade. Which is a fair demonstration of the blocking this PR fixes, but it means the test cannot double as its own control. It has the same exposure the two existing tests in that file already have.

## What I ran

macOS arm64, Mojo 1.0.0 (ed45d567), at `876eae1`.

| entry | result |
|---|---|
| `tests/http/test_server_ws_offload.mojo` | 3 passed |
| `tests/http/test_server_ws_upgrade.mojo` | 3 passed |
| `tests/runtime/test_thread_ffi.mojo` | 10 run, 10 passed |
| `tests/ws/test_ws.mojo` | 38 run, 38 passed |
| `tests/ws/test_ws_multicore.mojo` | 1 passed |
| `tests/http/test_server_multicore.mojo` | 4 run, 4 passed |
| `tests/_agg/agg_http` | 1235 run, 1234 passed, 1 failed |
| `tests/_agg/agg_ws` | 78 run, 78 passed |
| `tests/_agg/agg_runtime` | 205 run, 205 passed |
| `mojo format` | clean on all 8 changed files |

The `agg_http` failure is `test_stream_coalescing`, pre-existing and identical on `d282574`. Linux, the glibc stack default, `pin_to_cpu` and the `pthread_create` failure path are all still unverified from here.

Issue #5, opened 2026-07-22
HTTPS/h2/h3 Streamed Responses


bowyern:
I'm leaning a bit on GPT5.6 with this one. What would it take to bring streaming responses to https/h2/h3?

Happy to donate tokens if you have some guidance. I don't think I could audit the results in a reasonable time :D

ehsanmok:
Good question! let me think about it. Should be doable to some extent.

ehsanmok:
Added some support in main. [Here](https://github.com/ehsanmok/flare/blob/main/examples/advanced/https_server.mojo) is an example. Please let me know if you've any feedback.

bowyern:
I should have been more specific, I was looking for streaming on the HttpClient side. It looks like there's get_streaming_tls() but it downgrades to 1.1

ehsanmok:
Please check v0.10.0

bowyern:
It looks like a design was committed but not an implementation.

ehsanmok:
Reopened. Will check again.

bowyern:
I gave the committed plan to Sol xhigh and added streaming post for making OpenAI API streaming requests. It came up with a 3000 line PR. I'm breaking that into stacked PRs, but does that sound remotely reasonable?

ehsanmok:
Awesome! Definitely! I'll try to review as soon as I can.

ehsanmok:
Thanks for the PRs! the review and merge would be easier now that github supports stack PRs.

ehsanmok:
Done in 0.11.0. `HttpClient.get_streaming` is now scheme-agnostic and returns an `HttpStreamResponse` whose body you pull with `read_chunk`; the wire is whatever the handshake settles on, so HTTP/1.1, HTTP/2 over TLS and prior-knowledge h2c all work through the one call. `send_streaming(Request)` is the primitive underneath, and `post_streaming` covers the server-sent-events and JSON-lines shape. Uploads stream too: `send_chunked` takes a `body_size` for known-length bodies and a real `Cancel`, and a source that raises mid-body never emits the chunked terminator.

HTTP/3 has a reader (`Http3Download`) but no entry point wires to it yet, so h3 streaming is not reachable from `HttpClient` in this release. That is the one part of the original ask still outstanding.

Two runnable examples are in `examples/advanced/`: `http_stream_client.mojo` and `streaming_upload.mojo`. The interop smoke test pulls a megabyte through the new client on each wire.

[...]

PR #17, opened 2026-09-04
refactor(client): centralize TLS transport setup

bowyern:
## Summary

- move the owned TCP/TLS transport used by streaming HTTP/2 clients into a shared client module
- reuse that transport from gRPC streaming instead of maintaining a second implementation
- centralize client certificate and ALPN configuration in the common TLS context builder so every connection path applies the same policy

## Why

This is the transport-preparation layer for client streaming over HTTP/2 and HTTP/3. Keeping it separate makes the TLS behavior and ownership change reviewable without any response-streaming API changes.

Part of #5.

## Verification

- `tests/tls/test_tls.mojo` — 16 passed
- `tests/tls/test_tls_resume.mojo` — 3 passed
- `tests/grpc/test_grpc_streaming.mojo` — 6 passed
- Mojo formatter clean
- `git diff --check` clean

## Stack

Part 1 of 8; review this PR first. Next: #18.

All PRs in the stack target `main`, following the repository's existing contributor-stack convention. Later PRs therefore show cumulative diffs until their predecessors merge.

1. #17 — shared transport and TLS setup
2. #18 — incremental HTTP/1 reader
3. #19 — HTTP/2 bounds and flow control
4. #20 — HTTP/3 and QUIC hardening
5. #21 — cross-protocol response readers
6. #22 — public streamed-response API
7. #23 — streamed-upload API
8. #24 — documentation and focused test tasks


flare/tls/stream.mojo
```diff
@@ -394,6 +394,39 @@ def _build_ssl_ctx(imm lib: OwnedDLHandle, config: TlsConfig) raises -> Int:
...
+        if (
+            _do_ssl_ctx_load_cert_key(
+                lib, ctx, config.cert_file, config.key_file
+            )
+            != 0
+        ):
+            var err = _c_err(lib)
+            _do_ssl_ctx_free(lib, ctx)
+            raise TlsHandshakeError("mTLS cert/key load failed: " + err)
+    if len(config.alpn) > 0:
```
ehsanmok, line 408: This block moved up from `TlsStream.connect`, and the move is what I want long term. It also means `connect_timeout`, `connect_over_tcp` and `connect_resumed` start sending an ALPN extension, and none of the three ever sent one.

The visible one is `flare/http/client.mojo:1014`. `_send_https` reads `stream.alpn_selected()` at line 1017 and branches into `_send_h2_over_tls` at line 1019, and with nothing advertised the server could never pick `h2`, so that branch was dead on every real dial. From here on a plain `HttpClient.get("https://...")` comes back over HTTP/2 whenever the origin offers it, and `_connect_h2` at line 901 stops pooling those origins because it only pools when the negotiated protocol is not `h2`.

Nothing in `tests/` dials `connect_timeout` or `connect_over_tcp` with an ALPN list, so there is no coverage on either side of this. Please add one that asserts `alpn_selected() == "h2"` over `connect_timeout` against the local TLS server with `alpn = ["h2", "http/1.1"]`, and one over `connect_over_tcp`. Both fail on the old code.

flare/tls/stream.mojo
```diff
@@ -394,6 +394,39 @@ def _build_ssl_ctx(imm lib: OwnedDLHandle, config: TlsConfig) raises -> Int:
             _do_ssl_ctx_free(lib, ctx)
             raise TlsHandshakeError("Session cache setup failed: " + err)
 
+    # All client connection paths share identity and ALPN setup here.
+    if config.cert_file != "":
```
ehsanmok, line 398: `connect` guarded this with `cert_file != "" and key_file != ""`. Dropping the second half means a config carrying a cert with an empty `key_file` now reaches `_do_ssl_ctx_load_cert_key` and raises `mTLS cert/key load failed`, where before it went ahead with no client certificate.

I prefer the loud version, so I am not asking you to put the guard back. I do want it stated in the commit body, and a test that pins whichever contract you settle on.

flare/grpc/streaming.mojo
```diff
@@ -33,9 +33,9 @@ from ..http2.client import Http2ClientConnection
 from ..http2.hpack import HpackHeader
 from ..http.url import Url
 from ..net import NetworkError
-from ..runtime.pool import Pool
 from ..tcp import TcpStream
 from ..tls import TlsStream
+from ..http._client.h2_transport import _H2Transport
```
ehsanmok, line 38: `TcpStream` and `TlsStream` were only named by the struct you just moved out. Neither appears anywhere else in the file now, and the new import reads better next to the other `..http` one.

```suggestion
from ..http._client.h2_transport import _H2Transport
from ..http.url import Url
from ..net import NetworkError
```

flare/http/_client/h2_transport.mojo
```diff
@@ -0,0 +1,46 @@
+"""Owned TCP/TLS transport shared by the HTTP and gRPC stream readers."""
+
+from ...io.buf_reader import Readable
+from ...runtime.pool import Pool
+from ...tcp import TcpStream
+from ...tls import TlsStream
+
+
+struct _H2Transport(Movable, Readable):
```
ehsanmok, line 9: The docstring did not come along with the move. I am not usually asking for more comments, but "exactly one of `_tcp_addr` / `_tls_addr` is non-zero" is a constraint the types cannot carry and `read` on line 31 leans on it: the `_tcp_addr == 0` branch is only correct because a zero there means the TLS cell is the live one. Worth bringing back, with the post-`close()` state added since both addresses are zero then.

```suggestion
struct _H2Transport(Movable, Readable):
    """A live HTTP/2 transport: either a cleartext ``TcpStream`` or a
    TLS ``TlsStream``, stored in a heap cell (:class:`Pool`) so the
    move-only stream can be re-borrowed mutably for each read/write.

    Exactly one of ``_tcp_addr`` / ``_tls_addr`` is non-zero until
    :meth:`close`, after which both are zero. The cell is freed
    (closing the socket) by :meth:`close` or, as a backstop, by the
    destructor.
    """
```

flare/http/_client/h2_transport.mojo
```diff
@@ -0,0 +1,46 @@
...
+    var _tcp_addr: Int
+    var _tls_addr: Int
+    var _is_h2: Bool
```
ehsanmok, line 12: Set to `False` in the constructor and never read again in this layer. The first reader is `flare/http/client.mojo:1497` in #22, so it lands with a purpose later in the stack. I would rather it arrived with that PR than sit here as a field nothing in this PR touches, though I will not hold the layer over it if threading the constructor change through #18 to #21 is more churn than it is worth.

flare/http/_client/h2_transport.mojo
```diff
@@ -0,0 +1,46 @@
...
+
+    @staticmethod
+    def from_tcp(var s: TcpStream) raises -> _H2Transport:
+        return _H2Transport(Pool[TcpStream].alloc_move(s^), 0)
+
+    @staticmethod
+    def from_tls(var s: TlsStream) raises -> _H2Transport:
+        return _H2Transport(0, Pool[TlsStream].alloc_move(s^))
+
+    def read(mut self, buf: UnsafePointer[UInt8, _], size: Int) raises -> Int:
```
ehsanmok, line 31: Rebase item, and the textual merge will not flag it. `main` moved the `Readable` trait to `Pointer[UInt8, _]` on Sep 11 (`flare/io/buf_reader.mojo:40`) and took `TcpStream.read` and `TlsStream.read` with it (`flare/tcp/stream.mojo:402`, `flare/tls/stream.mojo:731`), and `Pool.get_ptr` returns a `Pointer` there too. This file declares `Readable` conformance with `UnsafePointer[UInt8, _]`, so the signature here and the two `get_ptr(...)[].read(...)` calls below need the same substitution once you rebase.

The copy still sitting in `flare/grpc/streaming.mojo` on `main` gets away with `UnsafePointer` because it does not declare the trait. Please compile this file after the rebase rather than trusting the merge to be clean.

flare/http/_client/h2_transport.mojo
```diff
@@ -0,0 +1,46 @@
...
+        if self._tcp_addr != 0:
+            return Pool[TcpStream].get_ptr(self._tcp_addr)[].read(buf, size)
+        return Pool[TlsStream].get_ptr(self._tls_addr)[].read(buf, size)
```
ehsanmok, line 34: After `close()` both addresses are zero, so a read here dereferences `Pool[TlsStream].get_ptr(0)` and takes the process down instead of raising. Carried over from the old code rather than introduced by you, and the gRPC callers drop the transport right after closing, so it is not reachable today.

It gets easier to reach in #22, where `HttpClient` hands this struct to a response object that can outlive the close. A `raise NetworkError(...)` when both addresses are zero costs nothing at this level. Happy for that to be a follow-up.

ehsanmok (changes requested):
Thanks for splitting the stack this way - putting the transport and the TLS setup in their own layer made this reviewable in one sitting, and the ownership side of `_H2Transport` holds up under inspection. I reproduced the four suites locally before reading the diff and they are green. What I can't take as-is is the label on it. Folding mTLS and ALPN into `_build_ssl_ctx` switches both on for three dial paths that never had them, and one of those changes what protocol `HttpClient` speaks for an ordinary `https://` request; the summary files that under a refactor. The summary's third bullet also says every connection path now applies the same policy, and `connect_resumed` still carries its own copy of the mTLS block, so that is not quite what landed either. I want both named in the commit message and the wire change covered by a test before this merges.

On the rebase: this predates the 2026-09-11 migration on `main` and conflicts only in `flare/tls/stream.mojo`, where the overlap is exactly the two places you touched, `_build_ssl_ctx` and `TlsStream.connect`. Main's other edits in that file (the `_do_ssl_*` shims and `TlsStream.read` / `write_all`) merge cleanly. There is one rebase item the textual merge will not catch, inline on the new file.

This layer is the bottom of the stack, so the test-wiring note the other seven point back at goes here. `main` now compiles tests per area into one binary under `tests/_agg`, driven by `pixi run tests` -> `tools/run_test_aggregates.sh`. The generated aggregate imports every module-level `def test_*` by name, matched by the `TEST_DEF` regex at `tools/gen_test_aggregates.py:95` across every file the glob at `:146` finds, so a test file that has not been regenerated into `tests/_agg` does not run at all. CI calls the generator with `--check` first (`tools/run_test_aggregates.sh:66`), so a stale aggregate fails the build rather than quietly dropping coverage. One `pixi run tests-gen` with the regenerated `tests/_agg` committed alongside is the whole chore. This layer adds no test file of its own; it is the two tests in item 1 that will need it.

## 1. Three dial paths now put ALPN on the wire that never sent it

Before this PR the mTLS and ALPN blocks lived inside `TlsStream.connect` and nowhere else. `connect_timeout`, `connect_over_tcp` and `connect_resumed` called `_build_ssl_ctx` and stopped there, so they ignored `TlsConfig.alpn` and `TlsConfig.cert_file` completely. Moving both blocks into the helper gives all four paths the same policy, which is where I want to end up. It is also a wire change for every caller of the other three:

| call site | ALPN before | ALPN after |
|---|---|---|
| `flare/http/client.mojo:897` `_connect_h2` | none | `h2`, `http/1.1` |
| `flare/http/client.mojo:1014` `_send_https` | none | `h2`, `http/1.1` |
| `flare/http/client.mojo:1012` `_send_https` via proxy | none | `h2`, `http/1.1` |
| `flare/http/client.mojo:1186` `_send_chunked_tls` | none | `http/1.1` |
| `flare/http/client.mojo:1444` `get_streaming_tls` | none | `http/1.1` |
| `flare/grpc/client.mojo:216` `GrpcClient._connect` | none | `h2` |

Row two is the one that changes user-visible behavior. `_send_https` branches on `stream.alpn_selected()` at `flare/http/client.mojo:1017`, and with nothing advertised the server had no way to select `h2`, so the `_send_h2_over_tls` call at line 1019 was unreachable on any real dial. After this change it is reachable, and `HttpClient.get("https://...")` against an h2-capable origin stops speaking HTTP/1.1. `_connect_h2` moves with it: line 901 only pools a connection when the negotiated protocol is not `h2`, so origins that now negotiate `h2` stop entering the TLS keep-alive pool and get re-dialled.

The gRPC row is a plain fix. `GrpcClient._connect` sets `alpn = ["h2"]` at `flare/grpc/client.mojo:214-215` and then dialled through `connect_timeout`, which threw the list away, so any server that requires ALPN for h2 was refusing it.

None of this has coverage. `tests/grpc/test_grpc_streaming.mojo` is cleartext h2c only, by its own module docstring at line 4, and its six tests all build a `GrpcClient` on an `http://` base. The four TLS suites dial through `TlsStream.connect`, which already applied ALPN, so they say nothing about the paths this PR changes. I'd like a test that dials `connect_timeout` against the local TLS server with `alpn = ["h2", "http/1.1"]` and asserts `alpn_selected()`, and the same over `connect_over_tcp`. Both fail on the old code, which is what makes them worth carrying.

## 2. `connect_resumed` still loads the client cert itself

`flare/tls/stream.mojo:985-995` kept its own copy of the mTLS block, under the comment "mTLS path matches connect()". By the time that block runs, `_build_ssl_ctx` has already loaded the same cert and key into the same `SSL_CTX`. OpenSSL takes the second load without complaint, so nothing breaks, but it is duplicated work and it undercuts the one claim in the PR title. Please delete lines 985-995.

## 3. The mTLS guard lost its `key_file` half

Inline on the added line. `cert_file != "" and key_file != ""` became `cert_file != ""`, so a config that sets a cert and leaves the key empty now raises `mTLS cert/key load failed` where it used to proceed without a client certificate. I think raising is the better contract. It is still a third behavior change riding in on a refactor, and it wants a line in the commit body plus a test pinning whichever way you land.

## 4. The commit message

One subject line and nothing under it. Items 1 and 3 are exactly what I want a body for here: which dial paths gain ALPN and client certificates, that an ordinary HTTPS request through `HttpClient` can now come back over HTTP/2, and what a caller who configured neither should expect. Rationale goes in the commit message in this repo, not into code comments.

## What I ran locally

Worktree at c5e1f96, Mojo 1.0.0 (ed45d567), macOS arm64.

| entry | result | wall |
|---|---|---|
| `tests/tls/test_tls.mojo` | 16 run, 16 passed, 0 failed | 4s |
| `tests/tls/test_tls_resume.mojo` | 3 run, 3 passed, 0 failed | 4s |
| `tests/grpc/test_grpc_streaming.mojo` | 6 passed | 19s |
| `tests/http/test_tls_client_pool.mojo` | 4 run, 4 passed, 0 failed | 4s |

`mojo format` is clean on all three changed files and `git diff --check` is clean.

What I could not verify: nothing in the tree exercises ALPN negotiation over `connect_timeout`, `connect_over_tcp` or `connect_resumed`, so the change in item 1 is uncovered on both sides of the diff and the green runs above do not speak to it. I also did not compile the branch rebased onto `main`.

Ownership is sound, and the fd closes exactly once. `_H2Transport.close()` dropping the explicit `.close()` calls and going straight to `Pool[T].free` is equivalent to the old shape: `Pool.free` runs `T`'s destructor at `flare/runtime/pool.mojo:111`, `TlsStream.__deinit__` at `flare/tls/stream.mojo:539` sends `close_notify` and frees the SSL objects, and the TCP fd is closed by `TcpStream.__deinit__` at `flare/tcp/stream.mojo:140`. `close()` zeroes both addresses and `Pool.free(0)` returns immediately, so the destructor that follows an explicit `close()` does nothing. Neither order double-closes.

No import cycle. `flare/http/_client/h2_transport.mojo` pulls `io.buf_reader`, `runtime.pool`, `tcp` and `tls`, and nothing under `flare/http`, `flare/tls`, `flare/tcp`, `flare/io` or `flare/runtime` imports `flare.grpc` at all, so the new direction of the dependency is safe.

The gRPC side lost nothing. `flare/grpc/streaming.mojo` has two hunks, the import swap and the struct deletion, so the status mapping, the deadline handling and the `h2` requirement are all untouched.

h3 and QUIC are not in the diff and do not route through `TlsStream`. The server side is untouched too: `flare/tls/acceptor.mojo` builds its own context and does its own ALPN at line 342, and it does not call `_build_ssl_ctx`.

## Minor

The header comment above the helper at `flare/tls/stream.mojo:356-362` still says it is shared by "`connect` and `connect_timeout`" and covers "new + security policy + verify mode + CA bundle". There are four callers now and the list is short by the session cache, mTLS and ALPN.

`flare/http/_client/__init__.mojo` enumerates the sub-modules in its docstring and `h2_transport` is missing from it.

## What I'll take

The two Minor items, in a follow-up, if you'd rather keep this layer to the transport, and the null-address guard in the inline comment on `read`, which can wait for a later PR.

From you: the two tests in item 1, the deletion in item 2, and a commit body.

ehsanmok:
Landed on \`main\` as c02f201, cherry-picked with your authorship intact.

Two follow-ups on top, both mine:

- 33ca8b2 restores the struct and member docstrings the move dropped, takes \`Pointer\` rather than the deprecated \`UnsafePointer\` in \`read\` (which is what \`Readable\` asks for now), and removes the \`_is_h2\` field, which was written once and never read at this layer.
- 98ad7cf adds tests for the part of this change that turned out to matter most. Folding mTLS and ALPN into \`_build_ssl_ctx\` does not just de-duplicate \`connect\` -- it fixes \`connect_timeout\` and \`connect_over_tcp\`, which set neither. Those are the only two paths \`HttpClient\` and the gRPC client ever take, so HTTP/2 over TLS was unreachable from the client regardless of what it advertised. Both new tests fail against the tree one commit before the fold.

Thank you. Closing since the commit is on main.

PR #18, opened 2026-09-04
feat(http): harden incremental HTTP/1 downloads

bowyern:
## Summary

- replace the HTTP/1 download path with an incremental body reader for content-length, chunked, bodyless, and EOF-delimited responses
- preserve duplicate headers and expose trailers without buffering the complete body
- reject conflicting or malformed framing, unsupported upgrades/transfer encodings, and oversized response heads, chunk lines, or trailers

## Why

This is the bounded HTTP/1 reader needed by the public streaming response API. The stricter framing rules also prevent ambiguous message boundaries and unbounded metadata growth on attacker-controlled responses.

## Stack

Part 2 of 8. **Stacked on #17; review and merge #17 first.**

Both PRs target `main`, following the repository's existing contributor-stack convention, so this PR's GitHub diff is cumulative until #17 merges. The unique commit for this layer is `2718065`.

## Verification

- `tests/http/test_download_decoder.mojo`
- Mojo formatter clean
- `git diff --check` clean


flare/http/_client/download.mojo
```diff
@@ -233,7 +281,7 @@ struct HttpDownload[R: Readable & Movable](Movable):
             )
             var size = _parse_hex(String(String(size_str).strip()))
```
ehsanmok, line 282: The `.strip()` here means `" 3"` and `"3 "` both parse as a chunk size of 3. The repo's other chunked decoder refuses that: `scan_chunked_end` at `flare/http/proto/chunked.mojo:143-150` fails on the first byte before the `;` that isn't a hex digit, and its `digits == 0` check covers the empty case that your new guard at `download.mojo:356` handles separately. Whitespace in front of a chunk size is a well-worn desync primitive, and since this PR is what made the chunk terminator strict at `:254`, the size line should match it.

None of the four cases in the new test file put whitespace in a size line, so dropping the strip doesn't disturb them.

```suggestion
            var size = _parse_hex(String(size_str))
```

flare/http/_client/download.mojo
```diff
@@ -50,60 +50,114 @@ struct HttpDownload[R: Readable & Movable](Movable):
...
+            raise NetworkError("HTTP download: invalid status code")
 
-        var content_length = -1
-        var is_chunked = False
         for li in range(1, len(lines)):
             var ln = lines[li]
             var colon = ln.find(":")
-            if colon < 0:
-                continue
+            if colon <= 0:
```
ehsanmok, line 159: Turning `colon < 0: continue` into a raise is the right move. It leaves the two field-line shapes that matter, though.

A line starting with SP or HTAB is a continuation, and here it has a colon at index > 0, so it becomes its own header field. `X-Foo: bar\r\n evil: value` yields a header named `evil`, where RFC 9112 §5.2 says the two lines fold into one value. `flare/http/proto/h1_leniency.mojo:104-110` has `allow_obs_fold` off by default for exactly this reason, and what happens here is worse than either accepting or rejecting, because the injected field is indistinguishable from a real one downstream.

The `.strip()` on the key just below has the same shape: `Content-Length : 5` parses as `content-length`, which `h1_leniency.mojo:96-102` names as the classic smuggling vector and turns off by default.

Both predate this PR so I won't hold the layer on them, but the fix is small and this is the hardening PR: reject the line when its first byte is SP or HTAB, and reject it when the byte before the colon is SP or HTAB instead of stripping.

flare/http/_client/download.mojo
```diff
@@ -50,60 +50,114 @@ struct HttpDownload[R: Readable & Movable](Movable):
...
+        if method.upper() == "HEAD" or self.status == 204 or self.status == 304:
+            self._done = True
+            self._buf = List[UInt8]()
+            return
+
+        var content_length = -1
+        var lengths = self.headers.get_all("content-length")
+        for i in range(len(lengths)):
+            var value = _parse_decimal(lengths[i])
+            if content_length >= 0 and value != content_length:
```
ehsanmok, line 127: This accepts repeated `Content-Length` when the values agree. The server side of the same tree takes the opposite default: `allow_multiple_content_length` at `flare/http/proto/h1_leniency.mojo:140-147` is off, and its docstring says the canonical treatment is 400 even when the values are identical.

Two parsers in one repo with opposite answers on the same field is what eventually bites someone. I'd like one answer picked. My preference is to match the server and reject duplicates outright, since a client that agrees with the origin about framing is worth more than one extra accepted response. If the lenient read is deliberate on the client side, then the constructor docstring should say so and that flag's docstring should record that the client differs.

flare/http/_client/download.mojo
```diff
@@ -50,60 +50,114 @@ struct HttpDownload[R: Readable & Movable](Movable):
...
+            self._parse_head(head)
+            if self.status == 101:
+                raise NetworkError(
+                    "HTTP download: protocol upgrades are not supported"
+                )
+            if self.status >= 200:
+                break
+            self.headers = HeaderMap()
+
+        if method.upper() == "HEAD" or self.status == 204 or self.status == 304:
```
ehsanmok, line 118: HEAD, 204 and 304 is the right set for a reader that only ever sees GET today. What's missing is a 2xx to CONNECT, which per RFC 9110 §9.3.6 also has no body, because the connection becomes a tunnel. As written such a response carries no CL and no TE, so it falls through to `_DL_MODE_CLOSE` at `:144` and the reader starts handing tunnel bytes back as though they were a body.

#21 refuses CONNECT at the caller (`flare/http/_client/stream_request.mojo:98`), so the streaming API is covered later in the stack. A direct `HttpDownload(stream^, "CONNECT")` still isn't, and this struct is exported from `flare/http/__init__.mojo:210`. A third arm on this condition closes it.

flare/http/_client/download.mojo
```diff
@@ -50,60 +50,114 @@ struct HttpDownload[R: Readable & Movable](Movable):
...
-    def __init__(out self, var stream: Self.R) raises:
+    def __init__(
+        out self,
+        var stream: Self.R,
+        method: String = "GET",
+        max_header_bytes: Int = 65536,
+        var initial: List[UInt8] = List[UInt8](),
+    ) raises:
         """Read + parse the response head from ``stream`` (request already
         written) and set up incremental body framing."""
```
ehsanmok, line 71: The constructor picked up `method`, `max_header_bytes` and `initial`, and this docstring still describes the old two-argument contract.

`max_header_bytes` is the one that needs saying out loud: it is no longer only the head limit, it also caps a single chunk-size line at `:262` and `:269` and the whole trailer section at `:308` and `:314`. Someone sizing it for headers alone would be surprised to find it rejecting a long chunk extension.

The module docstring at `:10-19` has the same problem. It still lists the three framing modes and says the chunked body is terminated by the `0` chunk, with nothing about trailers, informational heads, or any of the rejections this layer adds.

tests/http/test_download_decoder.mojo
```diff
@@ -0,0 +1,108 @@
...
+                + String(status)
+                + " Empty\r\nContent-Length: 123\r\n\r\n"
+            )
+        )
+        assert_equal(len(dl.read_chunk()), 0)
+    for framing in [
+        "Content-Length: -1",
+        "Content-Length: 4\r\nContent-Length: 3",
+        "Content-Length: 3\r\nTransfer-Encoding: chunked",
+        "Transfer-Encoding: gzip",
```
ehsanmok, line 75: `gzip` covers the unsupported-coding arm, but not the shapes that actually turn up in smuggling reports. I'd want `Transfer-Encoding: chunked, gzip` and `Transfer-Encoding: gzip, chunked` in this list, both rejecting since chunked has to be the final coding and you don't decode gzip. Then two separate `Transfer-Encoding: chunked` lines, and a `Content-Length: 4\r\nContent-Length: 4` case so whichever answer we settle on for duplicates-that-agree is pinned by a test rather than by reading the code.

While you're in here, the reject paths with no case anywhere in the file: 101 (`flare/http/_client/download.mojo:110`), a status outside 100-599 (`:153`), chunk-size hex overflow (`:370`), a `Content-Length` or `Host` trailer (`:331`), a trailer section over the bound (`:309`), a chunk-size line over the bound (`:262`), `max_header_bytes=0` (`:86`), the `initial` pre-buffer, and two informational heads in a row so the cumulative head bound at `:101-103` is exercised rather than just the single-head path.

The gap I care most about is EOF-delimited: a 200 with neither CL nor TE, some body, then EOF. That is the whole of `_DL_MODE_CLOSE` and nothing in this file touches it, including the part that matters, which is that it must not be reachable for a bodyless status.

ehsanmok (changes requested):
Thanks for this - the incremental reader is the right shape, and most of what the description promises is genuinely in the code. I read the layer commit against the tree and ran your test file plus the two existing streaming suites locally before writing this. Two things I want fixed before merge: the chunk-size line still takes whitespace around the hex, which is a desync primitive in a PR whose summary says it rejects malformed framing, and the ten or so new ways this layer refuses a response are almost entirely untested. After those, a short list I'm happy to carry myself.

On the rebase: this branch predates the 2026-09-11 migration on `main`, and the only file that conflicts for #18 is `flare/tls/stream.mojo`, inherited from #17. Run `pixi run tests-gen` after rebasing so `tests/http/test_download_decoder.mojo` joins `tests/_agg`; see my note on #17 for what that runner does.

## What I ran locally

| entry | result | summary |
|---|---|---|
| `tests/http/test_download_decoder.mojo` | pass | 4 tests run: 4 passed, 0 failed, 0 skipped (0.458 s) |
| `tests/http/test_client_stream_download.mojo` | pass | 3 passed |
| `tests/http/test_h1_trailers.mojo` | pass | 8 passed |
| `mojo format` | clean | 5 changed files |

No existing test was deleted or weakened. The diff adds one file and touches nothing else under `tests/`, and both streaming suites that already live on `main` still pass unchanged against the new reader.

What I could not check: `pixi run tests` end to end, since the branch has to be rebased first; and anything past the `Pieces` fake, so partial-read behaviour over a real socket is still only covered by the two older suites.

## What's right, so it doesn't get lost in the asks below

- Rejecting Transfer-Encoding alongside Content-Length is the same answer the server parser gives by default (`allow_te_chunked_when_cl_present`, `flare/http/proto/h1_leniency.mojo:149-157`). The precedence question is settled deliberately and consistently with the rest of the tree.
- `chunked, gzip` and `gzip, chunked` both fail the `len(encodings) != 1 or encodings[0].lower() != "chunked"` test at `flare/http/_client/download.mojo:132`, so chunked-not-final is covered even though the test only exercises bare `gzip`.
- The overflow guards at `download.mojo:369` and `:381` are right, and `_parse_decimal` rejects `+5`, empty, and non-digit values because the field value is stripped before it gets there.
- A truncated Content-Length body raises at `download.mojo:230` instead of returning short, which was the first thing I went looking for.
- Trailers live in their own `HeaderMap` and never merge into `headers`, and #21 keeps that separation at `flare/http/_client/stream_response.mojo:30-49` and `:89-101`. So duplicate preservation via `get_all` doesn't change what `header()` hands back.
- Requiring the chunk terminator on the next pull instead of blocking for it is a real improvement over `main`, which silently skipped the CRLF whenever it wasn't already buffered.

## 1. The reject paths need tests

The layer adds roughly ten new ways to refuse a response and four test functions, and most of the new branches have no case at all. None of it is hard to write, since the `Pieces` fake already does everything needed, but I don't want to merge attacker-facing reject logic that nothing exercises, and the house rule here is a test per behaviour change.

With no case anywhere in the file: 101 rejection (`download.mojo:110`), the status range check (`:153`), hex overflow (`:370`), the forbidden-trailer list (`:331`), the trailer size bound (`:309` and `:315`), the chunk-size line bound (`:262` and `:269`), `max_header_bytes <= 0` (`:86`), the `initial` pre-buffer, two informational heads in a row so the cumulative bound at `:101-103` gets exercised, and the whole EOF-delimited mode, which the new file never touches. Named suggestions inline.

## 2. The chunk-size line still accepts whitespace around the hex

`download.mojo:282` strips before parsing, so `" 3"` and `"3 "` both parse as a chunk size of 3. The repo's other chunked decoder refuses that: `scan_chunked_end` at `flare/http/proto/chunked.mojo:143-150` fails on the first byte before the `;` that isn't a hex digit. Whitespace in front of a chunk size is one of the standard desync primitives, and this is the PR that made the terminator on the other side of the same data strict, so the size line should match. One line, suggestion inline. It came over from `main` rather than being introduced here, but this is the change that should close it.

## 3. What I'll take as follow-ups unless you want them in this PR

All of these predate the diff. Each is a place where this parser is more lenient than `flare/http/proto/h1_leniency.mojo` says an H1 parser in this repo may be by default:

| relaxation | server default | this reader |
|---|---|---|
| `allow_obs_fold` (`:104`) | off, reject | accepts, and the folded line becomes a separate header |
| `allow_ows_around_colon` (`:96`) | off, reject | accepts, strips the whitespace |
| `allow_multiple_content_length` (`:140`) | off, reject | accepts when the values agree |
| `allow_lf_only_line_endings` (`:81`) | off, reject | accepts, via `_split_lines` at `flare/http/_client/parse.mojo:169` |

The obs-fold row is the one I care about. A continuation line is accepted and then reinterpreted as a distinct field, which is header injection, and none of the new validation catches it. Inline on the first two rows. If you'd rather keep this PR to the incremental reader I'll open an issue for the set and get out of your way.

## Minor

- The constructor picked up three parameters and the docstring at `download.mojo:70-71` mentions none of them. `max_header_bytes` in particular is no longer only the head limit. Inline.
- `download.mojo:308` checks the trailer bound before filling, and `_fill` pulls up to `_READ_BUF_SIZE` (16 KiB, `parse.mojo:21`), so the transient buffer can reach `max_header_bytes + 16 KiB`. The accepted total is bounded correctly at `:313-315`, so this is buffer shape rather than a bypass. Same pattern at `:262`.
- `:86` and `:215` raise a bare `Error` where every other failure in the file raises `NetworkError`, which changes the prefix a caller sees on the message.
- The commit message for this layer is the subject line and nothing else. For a change that settles framing precedence and adds a dozen reject paths, the reasoning has to live somewhere, and this repo puts it in the commit message rather than in comments.

## What I'll take

Everything in section 3, on our side, if you'd rather not widen the diff. Nothing in Minor blocks.

From you: the reject-path tests in item 1, the chunk-size line in item 2, and a commit body. Once those land I'll merge this, rebase included.

ehsanmok:
Landed on \`main\` as 7c4f2ee, cherry-picked with your authorship intact.

Two follow-ups, both mine:

- ea779d7 migrates the \`Pieces\` stub off the deprecated pointer spellings and regenerates \`tests/_agg\` so the four new tests actually run.
- 20e5740 closes the review items. Chunk-size lines no longer strip, so \`" 3"\` and \`"3 "\` stop meaning the same thing; obs-fold continuation lines and whitespace before a field-name colon are refused rather than reinterpreted; duplicate \`Content-Length\` is refused even when the values agree, matching the server side; and a 2xx to CONNECT is bodyless per RFC 9110 sec 9.3.6, which otherwise fell through to read-until-close and handed tunnel bytes back as a body. Six tests: three fail without those changes, three cover reject paths that already worked but had no case anywhere.

The strict framing in this PR is what made those four worth doing -- the terminator was already strict, so the size line was the remaining half. Thank you. Closing since the commit is on main.

PR #19, opened 2026-09-04
fix(http2): bound streaming flow control and request bodies

bowyern:
## Summary

- bound buffered HTTP/2 request bodies per stream and reset oversized requests without taking down unrelated streams
- enforce response ordering, bodyless-response, trailer, and `Content-Length` invariants
- tie streaming receive-window credit to application body drains, and discard pending uploads when a peer rejects or resets a stream

## Why

Streaming changes when flow-control credit can safely be returned. Returning credit as soon as bytes arrive permits unbounded buffering; withholding connection credit can stall unrelated streams. This layer keeps each response bounded by its stream window while allowing the rest of the connection to progress.

## Stack

Part 3 of 8. **Stacked on #18; review #17 and #18 first.**

The GitHub diff is cumulative against `main` until the preceding PRs merge. The unique commit for this layer is `5315218`.

## Verification

- `tests/http2/test_h2_config.mojo`
- `tests/http2/test_h2_state.mojo`
- `tests/http2/test_h2_streaming_state.mojo`


flare/http2/state.mojo
```diff
@@ -688,6 +716,65 @@ struct Connection(Copyable, Defaultable, Movable):
...
+        if self.is_client:
+            var status = 0
+            var regular = False
+            for field in hdrs:
+                if (
+                    field.name.byte_length() == 0
+                    or field.name != field.name.lower()
+                    or Connection._is_connection_specific(field.name)
+                ):
+                    raise Error("h2: invalid response field")
```
ehsanmok, line 728: This raise, and the ten more between here and `:776`, take the whole connection down for one bad response. `Connection.handle_frame` raises straight out through `Http2ClientConnection.feed`, and that method's docstring tells the caller to GOAWAY and close the socket, so a duplicate `:status` on stream 5 also kills streams 1, 3 and 7. RFC 9113 sec 8.1.1 says a malformed response "MUST be treated as a stream error of type PROTOCOL_ERROR".

This same function already gets that right three times: `:698-706` for the header-list cap, `:708-716` for the server-side request check, `:794-805` for the content-length mismatch. They all take the same shape, which is what I'd like here and at `:1197`:

```mojo
                    out.append(
                        self._rst_stream_frame(
                            sid, Http2ErrorCode.PROTOCOL_ERROR().value
                        )
                    )
                    s.state = StreamState.CLOSED()
                    self._put_stream(s^)
                    return out^
```

There's a second cost to raising from here. `s` is a copy that never reaches `_put_stream`, so the `s.recv_window -= len(f.payload)` at `:1186` goes with it, and a caller that catches and keeps going has flow-control accounting that no longer matches the peer's.

flare/http2/client.mojo
```diff
@@ -544,6 +588,9 @@ struct Http2ClientConnection(Defaultable, Movable):
         has no pending body."""
         if sid not in self._pending_body:
             return
+        if self.response_ready(sid) or sid in self._stream_errors:
+            self._discard_pending_body(sid)
```
ehsanmok, line 592: Dropping the body here is the right call, but nothing closes the stream out afterwards. No END_STREAM, no RST_STREAM, nothing returned to the caller. Our half stays open, so the peer keeps counting the stream against its own `SETTINGS_MAX_CONCURRENT_STREAMS`, and the request is truncated at whatever frame boundary we reached with no record of it anywhere.

`finish_upload` just below does exactly this properly, and its CLOSED / HALF_CLOSED_LOCAL check already covers the `sid in self._stream_errors` case, so it drops straight in:

```suggestion
            self.finish_upload(sid)
```

`test_h2_window_update_before_rejection_does_not_resume_upload` asserts `len(conn.drain()) == 0` after this path, which pins the silent version. That test's peer also sent RST_STREAM so the stream really is closed and there is nothing to send; the assertion would pass just as happily in the case where it isn't, which is the one that leaks.

flare/http2/state.mojo
```diff
@@ -1098,11 +1193,38 @@ struct Connection(Copyable, Defaultable, Movable):
...
+                raise Error("h2: DATA on a bodyless response")
+            if not self.is_client and len(body) > (
+                self.max_request_body_size - len(s.data)
+            ):
+                out.append(
+                    self._rst_stream_frame(
+                        sid, Http2ErrorCode.ENHANCE_YOUR_CALM().value
+                    )
+                )
+                s.data = List[UInt8]()
```
ehsanmok, line 1206: `s.headers` survives this, and nothing in the tree ever removes a CLOSED stream from `Connection.streams`. `_active_stream_count` at `:654` skips closed streams for the concurrency limit, so the peer isn't throttled either: it can open stream after stream, each with a header block up to the list cap and one oversized DATA frame, and keep every one of those header lists alive for the life of the connection. That turns an unbounded body buffer into unbounded header retention.

```suggestion
                s.data = List[UInt8]()
                s.headers = List[HpackHeader]()
```

Safe at this point, since `take_completed_streams` skips CLOSED streams and `take_request` is never reached for one. Pruning the slab properly pre-dates this PR and I'll take it separately.

flare/http2/client.mojo
```diff
@@ -399,11 +399,22 @@ struct Http2ClientConnection(Defaultable, Movable):
...
+            if len(self.inbox) >= 9:
+                var frame_size = (
+                    (Int(self.inbox[0]) << 16)
+                    | (Int(self.inbox[1]) << 8)
+                    | Int(self.inbox[2])
+                )
+                if frame_size > self.config.max_frame_size:
+                    raise Error(
+                        "h2 client: frame exceeds advertised maximum size"
+                    )
```
ehsanmok, line 413: RFC 9113 sec 4.2 makes an oversized frame a connection error of type FRAME_SIZE_ERROR, which means a GOAWAY carrying that code so the peer learns why. Raising closes the socket with nothing on the wire. `feed`'s own docstring does tell the caller to GOAWAY on a raise, so this is consistent with what's already in the method, and I'm happy to take the code-carrying version as a follow-up.

The part I would fix here is that the inbox is never advanced past the offending header, so a caller that catches and keeps feeding raises on every subsequent call for the life of the connection. The loop reads as if it recovers. Either drain the inbox before raising or say in the docstring that the connection is finished.

flare/http2/client.mojo
```diff
@@ -440,20 +451,53 @@ struct Http2ClientConnection(Defaultable, Movable):
...
+                ) & 0x7FFFFFFF
+                var code = (
+                    (Int(frame.payload[4]) << 24)
+                    | (Int(frame.payload[5]) << 16)
+                    | (Int(frame.payload[6]) << 8)
+                    | Int(frame.payload[7])
+                )
+                for entry in self.conn.streams.items():
+                    if entry[0] > last or code != 0:
+                        self._stream_errors[entry[0]] = code
```
ehsanmok, line 482: On a graceful GOAWAY (`code == 0`), every stream above `last` lands in `_stream_errors` with the value 0. Those streams were refused: the peer is telling us it never processed them and they are safe to retry on a fresh connection. Recording 0 makes `stream_error(sid).value() == 0` for them, which is indistinguishable from the clean RST_STREAM(NO_ERROR)-after-a-complete-response case you just taught `:455-463` to recognise.

REFUSED_STREAM (7) is the code that carries the retry-safe meaning. I'd record that for the `entry[0] > last` streams and keep `code` for the `code != 0` ones. Worth a test too, since the retry logic in #22 is going to read this.

flare/http2/client.mojo
```diff
@@ -1091,5 +1187,24 @@ struct Http2ClientConnection(Defaultable, Movable):
         try:
             if sid in self.conn.streams:
                 _ = self.conn.streams.pop(sid)
+            self._discard_pending_body(sid)
         except:
             pass
+
+    def initial_response_headers(self, sid: Int) raises -> List[HpackHeader]:
```
ehsanmok, line 1194: These two and `cancel_stream` below are new public API with no docstrings, in a file where every other public method has one.

Worth lining them up with the neighbour while you're here: `response_headers` at `:1175` returns an empty list when `sid` isn't in `conn.streams`, while these index the dict directly and raise. A caller reading trailers after `discard_stream` at `:1184`, or after `take_response` pops the stream, gets a `KeyError` from one accessor and `[]` from the other for the same input.

tests/http2/test_h2_streaming_state.mojo
```diff
@@ -0,0 +1,283 @@
...
+    conn.enable_response_streaming(1)
+    var encoder = HpackEncoder()
+    _h2_head(conn, encoder, [HpackHeader(":status", "200")])
+    _ = conn.drain()
+
+    _h2_data(conn, 3, padding=2)
+    assert_equal(conn.conn.streams[1].recv_window, 65532)
+    var updates = conn.drain()
+    assert_equal(_h2_credit(updates, 1), 3)  # Pad Length + padding only.
+    assert_equal(_h2_credit(updates, 0), 6)  # Other streams can progress.
```
ehsanmok, line 159: This comment is the one claim in the PR I couldn't get the tests to back up. The 6 bytes are stream 1's own connection-level credit, so what the assertion shows is that connection credit is returned in full for the frame we just received. That it lets a *different* stream keep going is the headline of the description, and nothing here exercises it.

A second stream on this connection, never drained, with stream 1 still receiving and completing, is about fifteen lines on top of what you already have and would close the gap.

ehsanmok (changes requested):
Thanks for this - the three problems this layer picks out are real ones, and `received_body_bytes` is the fix I would have missed. Once `drain_body` starts emptying `s.data`, the old `len(s.data) != content_length` comparison would have fired on every drained streaming response, so that counter is load-bearing. I reproduced the h2 area locally at `5315218` before reviewing and it is green; numbers below. I can't merge it as it stands. Two of the new response checks let one malformed response take down every other stream on the connection, which is the opposite of what the description promises, and one path drops a request body without telling anybody.

On the rebase: this conflicts with `main` only in `flare/tls/stream.mojo`, inherited from #17. `flare/http2/client.mojo`, `server.mojo` and `state.mojo` all auto-merge, but after the 2026-09-11 API migration a clean textual merge is not a clean compile, so I'd like to see it build on current `main` whatever happens with the items below. Run `pixi run tests-gen` after rebasing so `tests/http2/test_h2_streaming_state.mojo` joins `tests/_agg`; see my note on #17 for what that runner does.

## What I ran

At `5315218` in a clean worktree, Mojo 1.0.0 (ed45d567), macOS arm64:

| entry | result |
|---|---|
| `tests/http2/test_h2_config.mojo` | 18 passed, 0 failed |
| `tests/http2/test_h2_streaming_state.mojo` | 8 passed, 0 failed |
| `tests/http2/test_h2_state.mojo` | 14 passed |
| `tests/http2/test_h2_client_conn.mojo` | 11 passed |
| `tests/http2/test_h2_server.mojo` | 9 passed |
| `tests/http2/test_h2_per_stream_cancel.mojo` | 6 passed |
| `mojo format` | clean on the 11 changed files |

I did not run h2spec myself, so I'm taking CI's word on conformance. I also did not drive the streaming path over a real socket, because the driver that calls `enable_response_streaming` only shows up in #21. Everything below comes from reading the state machine and the new unit tests.

One warning in my logs that isn't yours: `flare/http2/server.mojo:551` still reaches `Http2ErrorCode` implicitly through the package `__init__`. `main` has already migrated those sites, so it should go away on rebase.

## 1. The new response checks raise, and a raise takes the connection with it

`flare/http2/state.mojo:728` through `:776` adds eleven `raise Error("h2: ...")` sites for response header validation, and `:1197` adds a twelfth for DATA on a bodyless response. `Connection.handle_frame` raises straight out through `Http2ClientConnection.feed`, whose own docstring tells the caller to send GOAWAY and close the socket. So a server that sends a duplicate `:status` on stream 5, or an uppercase field name, or a `content-length` in trailers, or one DATA byte after a 204, kills streams 1, 3 and 7 along with it.

RFC 9113 sec 8.1.1 puts malformed requests and responses at stream scope: they "MUST be treated as a stream error of type PROTOCOL_ERROR". The same function already does that three times over - the header-list cap at `:698-706`, the server-side request validation at `:708-716`, and the content-length mismatch at `:794-805` each emit `_rst_stream_frame`, close the stream and return. The client-side response checks are the only ones that don't, and the description's "without taking down unrelated streams" holds for the body bound alone.

A raise costs you two more things here. `s` is a copy that never reaches `_put_stream`, so the `s.recv_window -= len(f.payload)` at `:1186` is thrown away with it, and a caller that catches and carries on has stream accounting that has silently diverged from the peer's. And no error code reaches the peer, so the server sees a dropped socket with no reason attached.

`flare/http2/state.mojo:1169-1170` is a milder version of the same thing. DATA before the final response HEADERS is a malformed message under sec 8.1.1, so it belongs at stream scope too, and `_conn_error` sends GOAWAY. `test_h2_rejects_data_before_final_response_headers` asserts `goaway_sent`, so the test currently pins the behavior I'm asking you to change.

I'd like all of these converted to the `_rst_stream_frame(sid, PROTOCOL_ERROR)` shape used three lines up, with a test showing that a sibling stream on the same connection survives one malformed response. Suggestion inline.

## 2. `pump_pending_body` drops an upload without ending or resetting the stream

`flare/http2/client.mojo:591-593`. When a WINDOW_UPDATE arrives and the response is already ready, the stashed request body is dropped and nothing else happens: no END_STREAM, no RST_STREAM, nothing handed back to the caller. Our half of the stream stays open, so the peer goes on counting it against its own `SETTINGS_MAX_CONCURRENT_STREAMS`, and the request is truncated at whatever frame boundary we happened to reach with no record that it was truncated.

`finish_upload`, two functions down, is the behavior this wants, and it already guards the reset case with its CLOSED / HALF_CLOSED_LOCAL check. One-line fix, inline.

One caveat about `finish_upload` itself, recorded here because this is the layer that adds it. It half-closes with an empty DATA plus END_STREAM (`flare/http2/client.mojo:626`), where RFC 9113 section 8.1 has the client send RST_STREAM with NO_ERROR after an early reply. The difference is that our framing makes a deliberately truncated request body look complete to the server. It does not change the ask above, since an honest half-close beats a silent drop, and on the one path that uses it today (#23's `send_chunked`) the connection closes straight after so nothing observes it. I'd rather the framing were settled here than the first time someone reuses `finish_upload` on a connection that survives the exchange.

`test_h2_window_update_before_rejection_does_not_resume_upload` asserts `len(conn.drain()) == 0` after this path, which pins the silent drop. In that test the peer also sent RST_STREAM, so the stream really is closed and there is genuinely nothing left to send; the same assertion would pass in the case where it isn't closed, which is the case that leaks.

## 3. The bound frees the body and keeps the headers

`flare/http2/state.mojo:1198-1211`. ENHANCE_YOUR_CALM is defensible for this. REFUSED_STREAM would be wrong because we have already processed the request, and FLOW_CONTROL_ERROR would be a lie since the peer stayed inside its window. The stream-scoped reset with no GOAWAY is right, and `test_request_body_limit_resets_only_oversized_stream` shows stream 3 completing while stream 1 is reset, which is the part of the description I most wanted proved.

What the path doesn't release is `s.headers`, and nothing in the tree ever removes a CLOSED stream from `Connection.streams`. `_active_stream_count` at `:654` skips them for the concurrency limit, so the peer isn't throttled either. A peer can open stream after stream, each with a header block up to the list cap plus one oversized DATA frame, and keep every one of those header lists alive for the life of the connection. Clearing the headers is one line, inline; pruning the slab is the real fix and pre-dates this PR, so I'll take that separately.

While I'm in there: the bound reads `len(body) > self.max_request_body_size - len(s.data)`, using `s.data` rather than the `received_body_bytes` counter you just added. They agree on the server today because nothing drains `s.data` before END_STREAM, but the counter is the one that survives a drain, and using it keeps the two length checks in this function consistent with each other.

## 4. Flow control: what checks out, and what still isn't covered

The starvation question in the description checks out. Connection credit is returned in full at `flare/http2/state.mojo:1252` on every DATA frame regardless of `defer_body_credit`, and only the stream-level increment at `:1251` is withheld until `drain_body`, so a stalled reader on stream 1 cannot stop stream 3. The arithmetic holds as far as I can follow it. `pending_body_credit` only ever restores bytes already debited at `:1186`, so `recv_window` cannot climb past its initial value, which also puts the `& 0x7F` mask on the top byte of `_window_update_frame` out of reach. Both `drain_body` and the DATA path guard `credit > 0` before emitting, so no zero-increment WINDOW_UPDATE goes out. And deferred credit is never held past the stream's life: `drain_body` skips the update once `data_complete` or CLOSED, and the connection's share was returned at receive time.

The deadlock is a documented contract rather than a bug. `enable_response_streaming`'s docstring says to drain outbound WINDOW_UPDATEs before waiting for more inbound data, and a caller that reads before draining `outbox` will hang once the window closes. The only caller lands in #21, so I won't argue it here, but it wants a test at the level where the reader lives.

Cases I'd like covered, in the order I want them:

- Two streams on one connection, one never drained, showing the other still receives and completes. This is the headline claim and nothing currently exercises it.
- A malformed response with a sibling stream on the same connection, per item 1. Duplicate `:status` is the cheapest to construct.
- `content-length` longer than the DATA delivered. `test_h2_truncated_content_length_is_a_stream_error` covers the short direction; the long one goes through the same comparison and nothing pins it.
- 204 or 304 followed by DATA. `response_body_allowed` is set from two places and only the HEAD one is tested.
- `max_body_size = 0`, since the docstring says it permits only empty bodies and the check at `:1198` is a subtraction I'd like nailed down.

The last two I'll take as a follow-up if you'd rather not grow this PR further. The first two I'd like here, because they are the assertions the description makes.

## 5. Commit message

`5315218` is a subject line with an empty body. This layer also adds a configuration knob, four new public methods on `Http2ClientConnection`, a frame-size precheck in `feed`, GOAWAY-driven stream-error propagation, and a HALF_CLOSED_REMOTE to CLOSED transition in `send_data`. None of those appear in the subject or in the three bullets of the description. Rationale lives in the commit message in this repo, so please move the "Why" paragraph there and list the behavior changes while you're at it.

## Minor

`initial_response_headers` and `response_trailers` are new public API with no docstrings, and they raise on an unknown `sid` where the neighbouring `response_headers` returns an empty list. Inline.

## What I'll take

- Pruning CLOSED streams out of `Connection.streams` on the server, which pre-dates this PR.
- The `max_body_size = 0` and 204/304-with-DATA tests, if you'd rather not carry them here.
- The frame-size precheck's GOAWAY code, if you'd sooner not touch `feed` again in this layer; see the inline.

What I need from you is items 1 and 2, the first two tests in item 4, and the commit message. The rest is mine or optional.

ehsanmok:
Landed on \`main\` as 17b2dab, cherry-picked with your authorship intact. One conflict, three lines: \`main\` had dropped the redundant \`Movable\` and added \`@fieldwise_init\` to \`Http2Config\`, so I kept that and took your docstring.

Two follow-ups, both mine:

- 8de0ec2 regenerates \`tests/_agg\` for the streaming-state tests.
- 97cf6a2 closes the review items. The eleven response-head checks are stream errors now rather than raises, per RFC 9113 sec 8.1.1 -- a raise out of \`feed\` means the caller GOAWAYs, so a duplicate \`:status\` on stream 5 was killing streams 1, 3 and 7. That also preserves the \`recv_window\` decrement the raise discarded. The oversized-body reset drops the header list as well as the body, since nothing removes a CLOSED stream from the table and \`_active_stream_count\` skips them, so the peer was not throttled either. \`pump_pending_body\` calls \`finish_upload\` instead of silently dropping. And a graceful GOAWAY records REFUSED_STREAM for streams above \`last_stream_id\` rather than 0, which was indistinguishable from a clean reset after a complete response.

Two new tests: a malformed head on one stream leaves a sibling completable, and a graceful GOAWAY marks unprocessed streams retry-safe. Both fail before.

The deferred-credit mechanism here is what makes bounded-memory streaming possible at all; it is the layer the client API will sit on. Thank you. Closing since the commit is on main.

PR #20, opened 2026-09-04
fix(http3): isolate stream failures and recover control frames

bowyern:
## Summary

- make the HTTP/3 response reader incremental while enforcing final-header, trailer, bodyless-response, field-size, and `Content-Length` rules
- isolate reset and parse failures to the affected request so responses delivered for other multiplexed streams are retained
- add QUIC send-capacity accounting, receive-credit release, cancellation, and retransmission of flow-control/control frames
- bound upload waits while pumping peer flow-control updates

## Why

A malformed or stalled HTTP/3 stream must not corrupt another response or grow memory without a bound. QUIC control frames also need normal loss recovery; otherwise cancellation or newly granted credit can disappear after packet loss and deadlock a stream.

## Stack

Part 4 of 8. **Stacked on #19; review #17–#19 first.**

The GitHub diff is cumulative against `main` until the preceding PRs merge. The unique commit for this layer is `20ff904`.

## Verification

- `tests/h3/test_h3_client_mux.mojo`
- `tests/h3/test_h3_response_reader.mojo`
- `tests/quic/test_quic_client.mojo`


flare/http3/client.mojo
```diff
@@ -359,42 +366,142 @@ struct Http3ClientConnection(Movable):
...
+                                break
+                    except e:
+                        pr.reader._fail(String(e))
+            if events.connection_closed and not pr.reader.is_complete():
+                pr.reader._fail("connection closed before response FIN")
+            if pr.reader.has_error():
+                pr.reasm.pending = Dict[UInt64, List[UInt8]]()
+                pr.reader.body = List[UInt8]()
+                pr.reader.inbox = List[UInt8]()
+                credits.append(UInt64(0))
```
ehsanmok, line 484: Releasing zero here is right for bytes we are discarding, but nothing else happens either: the entry stays in `_pending` for the life of the connection, the skip at the top of the loop means we never read another byte of that stream, and the peer is never told. Its `max_recv_data` stops moving while `recv_offset` keeps climbing, and when it passes, `flare/quic/state.mojo:347-350` raises a flow-control violation that `flare/quic/client.mojo:861` swallows by dropping the entire packet, healthy streams' frames and the ACK included. With the default 1 MiB `initial_max_data` (flare/quic/state.mojo:281) a server mid-body reaches that quickly, and the isolation this PR is about stops holding.

`cancel_stream` (flare/quic/client.mojo:1363) is the piece that closes it and has no caller in this layer. On the first transition into the error state, send STOP_SENDING once and drop the request, something like:

```mojo
            if pr.reader.has_error():
                if sid not in self._cancelled:
                    self.quic.cancel_stream(sid)
                    self._cancelled[sid] = True
```

keeping the accessors raising off a retained error string rather than off `_pending`. Whatever shape you pick, I would like a test where the peer keeps writing on the failed stream past the window and a second request on the same connection still completes.

flare/http3/client.mojo
```diff
@@ -359,42 +366,142 @@ struct Http3ClientConnection(Movable):
...
+        """
+        var pos = 0
+        while pos < len(body):
+            var count = min(16384, len(body) - pos)
+            var wire = List[UInt8]()
+            encode_request_data(Span(body)[pos : pos + count], wire)
+            self._send_request_wire(sid, wire, timeout_ms)
+            pos += count
+        if fin:
+            self.quic.send_stream(sid, List[UInt8](), True)
```
ehsanmok, line 422: The FIN always rides its own packet here, so a GET costs two ack-eliciting datagrams where `send_request` sent one, and a body costs one more than it needs. `_send_request_wire` already knows which write is the last one; threading `fin` down and passing it on the final chunk keeps this to a single call, with the empty-body case falling back to the zero-length FIN send.

flare/http3/client.mojo
```diff
@@ -359,42 +366,142 @@ struct Http3ClientConnection(Movable):
...
+        var pos = 0
+        var start = _monotonic_ms()
+        while pos < len(wire):
+            self._check_stream_error(sid)
+            var cap = self.quic.send_capacity(sid)
+            if cap == 0:
+                if timeout_ms > 0 and _monotonic_ms() - start >= UInt64(
+                    timeout_ms
+                ):
+                    raise Error("h3 client: upload flow-control timeout")
```
ehsanmok, line 436: The bound is sane, since `start` resets after every chunk that goes out, so this is 30 s of no progress rather than 30 s per request. What is missing is the frame that tells the peer why we are quiet. RFC 9000 §4.1 wants STREAM_DATA_BLOCKED (and DATA_BLOCKED when the connection limit is the binding one) when we have data and no credit, and we send neither.

It matters more than it looks, because `derive_peer_send_limits` defaults a missing `initial_max_stream_data_bidi_remote` to 0 (flare/quic/transport_params.mojo:218-223). Against a peer that grants credit on demand, `send_capacity` is 0 from the first byte, nothing we send prompts an update, and the request dies after 30 s with an error that reads like the peer was slow. `main` failed the same case immediately with the limit in the message. Emitting the blocked frame once per stall, when `cap == 0` and we have not already sent one at this offset, covers both.

flare/quic/client.mojo
```diff
@@ -1342,40 +1360,106 @@ struct QuicClientConnection(Movable):
...
+        """
+        var final_size = UInt64(0)
+        if stream_id in self.send_offsets:
+            final_size = self.send_offsets[stream_id]
+        var payload = List[UInt8]()
+        encode_stop_sending(StopSendingFrame(stream_id, UInt64(0x10C)), payload)
+        encode_reset_stream(
+            ResetStreamFrame(stream_id, UInt64(0x10C), final_size), payload
+        )
+        self._send_padded_1rtt(payload^, ack_eliciting=True)
```
ehsanmok, line 1374: After this returns we have sent RESET_STREAM but changed nothing locally, so a later `send_stream` on the same id still emits STREAM frames, which RFC 9000 §3.1 forbids once the sender has reset. `send_offsets[stream_id]` also keeps advancing, so a PTO retransmit of this RESET_STREAM would carry a final size different from the one the peer first saw, which is a FINAL_SIZE_ERROR on their side. Marking the stream before the send closes both:

```mojo
        if stream_id in self.conn.streams:
            var s = self.conn.streams[stream_id]
            s.state = STREAM_STATE_RESET_SENT
            self.conn.streams[stream_id] = s
        self._send_padded_1rtt(payload^, ack_eliciting=True)
```

with `send_stream` raising on a stream in that state (`STREAM_STATE_RESET_SENT` is already exported from `flare/quic/state.mojo`; this import would need adding alongside `STREAM_STATE_RESET_RECVD`). It becomes reachable in #21, where `H3Stream.close` cancels while an upload loop may still be running on the same connection.

flare/quic/client.mojo
```diff
@@ -1342,40 +1360,106 @@ struct QuicClientConnection(Movable):
...
+        self.conn.max_data_recv += consumed
+        self.conn.streams[stream_id] = s
+        var payload = List[UInt8]()
+        encode_max_stream_data(
+            MaxStreamDataFrame(stream_id, s.max_recv_data), payload
+        )
+        encode_max_data(MaxDataFrame(self.conn.max_data_recv), payload)
+        # Credit updates must survive loss even when the peer has no credit
+        # left to send another STREAM frame and trigger a fresh update.
+        self._send_padded_1rtt(payload^, ack_eliciting=True)
```
ehsanmok, line 1425: The comment is right that these have to survive loss, and tracking them is the correct call. My worry is the rate: `poll_responses` calls this once per stream per poll whenever any ordered byte was delivered, so a bulk download emits one extra ack-eliciting datagram per burst per stream, each of which the peer then has to acknowledge and we have to hold in `_loss.sent` until it does. RFC 9000 §4.2 has the usual guidance of updating when the remaining window falls below a fraction of the whole rather than per chunk.

A counter of bytes consumed since the last update, flushed when it crosses half the window, gives the same liveness with a bounded packet count. Fine as a follow-up if you would rather land the correctness part first, but it should not ship as one packet per drain.

flare/quic/client.mojo
```diff
@@ -1342,40 +1360,106 @@ struct QuicClientConnection(Movable):
...
+
+    def release_stream_credit(
+        mut self, stream_id: UInt64, consumed: UInt64
+    ) raises:
+        """Replenish receive credit for newly consumed, ordered stream bytes."""
+        if consumed == 0 or stream_id not in self.conn.streams:
+            return
+        var s = self.conn.streams[stream_id]
+        s.max_recv_data += consumed
+        self.conn.max_data_recv += consumed
```
ehsanmok, line 1416: `conn.max_data_recv` is doing two jobs. Here it is the connection-level receive limit we advertise, which is what you want. It is also the seed for every stream created afterwards, both at `flare/quic/state.mojo:344` and in the new block at `flare/quic/client.mojo:1321`, so after 4 MiB of drained body on one request, the next stream opens with `max_recv_data` at 5 MiB rather than the per-stream limit we actually advertised in our transport parameters. That is the limit `apply_stream` enforces at `flare/quic/state.mojo:347`, so the per-stream check loosens with unrelated traffic and stops matching what the peer was told.

Keeping a separate `initial_max_stream_data` field for the seed and leaving `max_data_recv` purely connection-level would keep the two apart.

flare/http3/response_reader.mojo
```diff
@@ -317,18 +412,52 @@ struct Http3ResponseReader(Copyable, Movable):
...
             self._fail(String("response HEADERS missing :status"))
+            return
+        if self.status == 101:
+            self._fail("101 is invalid in HTTP/3")
+            return
+        if self.status < 200:
+            self.headers = List[QpackHeader]()
+            self._content_length = -1
+            self.status = 0
+            return
```
ehsanmok, line 446: Dropping the interim headers is right, but `_header_bytes` keeps the 1xx section's bytes on the books, and so do the trailers later. SETTINGS_MAX_FIELD_SECTION_SIZE is a per-section limit in RFC 9114 §4.2, and a couple of large 103 Early Hints followed by an ordinary final head can trip the 64 KiB default with no single section anywhere near it. Resetting the counter with the rest of the interim state:

```mojo
        if self.status < 200:
            self.headers = List[QpackHeader]()
            self._content_length = -1
            self.status = 0
            self._header_bytes = 0
            return
```

The same argument applies to the trailer section; charging it to a fresh budget would want a reset in the trailer branch of `_consume_frame` too.

tests/quic/test_quic_client.mojo
```diff
@@ -157,8 +158,53 @@ def test_client_handshake_through_retry() raises:
...
+                    break
+            except:
+                break
+        client._loss.sent[0].time_ms = 1
+        client._check_pto()
+        assert_equal(client.tx_1rtt_pn, pn + 2)
+        assert_equal(client._loss.pto_count, 1)
+        assert_equal(client._loss.outstanding(), 1)
+        assert_equal(client._loss.sent[0].pn, pn + 1)
+        assert_equal(client._loss.sent[0].frames, original)
```
ehsanmok, line 200: This assertion pins the retransmission to the exact bytes of the original, which is what `fire_pto` does today and is correct for RESET_STREAM. It is the wrong contract for MAX_STREAM_DATA and MAX_DATA. RFC 9000 §13.3 says to send the current limit when one of those is lost, and a byte-identical retransmit re-sends a value the peer already has, so if the newest update was the one that went missing, the peer stays blocked until something else moves. Worth a comment here saying the frames are opaque to the tracker, and worth the case the suite is missing: release credit twice, lose the first packet, and assert the probe carries the later limit.

The other gap on the QUIC side is inbound STOP_SENDING. `apply_stop_sending` flips the stream to RESET_SENT and no RESET_STREAM ever goes out (RFC 9000 §3.5), which no test would currently catch.

tests/h3/test_h3_response_reader.mojo
```diff
@@ -0,0 +1,70 @@
...
+    encode_response_headers(200, [QpackHeader("content-length", "12345")], wire)
+    reader.feed(Span(wire))
+    reader.signal_fin()
+    assert_false(reader.has_error())
+    var missing = Http3ResponseReader()
+    missing.signal_fin()
+    assert_true(missing.has_error())
+
+
+def main() raises:
```
ehsanmok, line 69: Three tests for the number of rules the reader gained, and each one bundles several assertions, so a failure will not say much about which rule broke. The partial-DATA test in particular is doing four jobs at once.

The reader takes no socket, so the negative cases are cheap. Before merge I would like: DATA total over and under `Content-Length`; a 204 or 304 carrying a non-empty DATA frame; SETTINGS or PUSH_PROMISE on a request stream; and a reserved or grease type with a large declared length, asserting `len(reader.inbox) == 0` so the incremental skip is pinned the way the DATA path already is here.

Also worth splitting the informational case out of `test_h3_informationals_trailers_and_length_after_draining`, since it currently proves 103 is dropped and trailers work and Content-Length matches, and any one of those failing reads the same.

ehsanmok (changes requested):
Thanks for this - it is the layer of the stack I was most worried about, and most of it lands. The response reader is genuinely incremental now, `_StreamReasm` rejects an inconsistent QUIC final size instead of believing the last FIN it sees, and `_check_send_limits` tracks MAX_DATA / MAX_STREAM_DATA instead of pinning to the initial values. The module-docstring rewrite is an accuracy fix too: the old text claimed one STREAM frame per `send_stream` call with oversized bodies rejected, and `send_stream` on `main` already loops over the MTU, so the old wording described a version of the code that is not there. I reproduced the three suites on your head before reading the diff and they are green.

What I cannot merge yet is the isolation story. A request whose reader fails is left holding receive credit that never comes back, nothing ever tells the peer to stop, and the remedy is `cancel_stream`, which this PR adds and never calls. Details in item 1, and the rest are smaller.

On the rebase: this conflicts with `main` only in `flare/tls/stream.mojo`, inherited from #17. `flare/http2/{client,server,state}.mojo` and `flare/quic/client.mojo` auto-merge textually, but they predate the deprecated-API migration, so a clean merge is not the same as a clean build here. Run `pixi run tests-gen` after rebasing so `tests/h3/test_h3_response_reader.mojo` joins `tests/_agg`; see my note on #17 for what that runner does.

## What I ran

Worktree at `20ff904`, Mojo 1.0.0 (ed45d567), macOS arm64.

| entry | result | wall |
|---|---|---|
| `tests/h3/test_h3_client_mux.mojo` | `test_h3_client_mux: 2 passed` | 13 s |
| `tests/h3/test_h3_response_reader.mojo` | `3 tests run: 3 passed , 0 failed , 0 skipped` | 5 s |
| `tests/quic/test_quic_client.mojo` | `test_quic_client: 4 passed` | 10 s |
| `mojo format` | clean over 17 changed files | |

What I did not run: the rebase build, the h3 and quic areas as a whole, and anything against a server other than flare's own. Everything I say below about flow control is read off the code, not measured on the wire, so correct me if you have a capture that says otherwise.

## 1. A failed request never gives its credit back and never cancels its stream

In `poll_responses`, a reader that has failed gets `credits.append(UInt64(0))` (flare/http3/client.mojo:484), keeps its entry in `_pending` forever, and has its chunks skipped on every later poll. Nothing goes out on the wire. The server has no reason to stop, `apply_stream` keeps advancing `recv_offset` against a `max_recv_data` that no longer moves, and when it passes it, `flare/quic/state.mojo:347-350` raises a flow-control violation. That raise unwinds into the `except: return` at `flare/quic/client.mojo:861`, which drops the whole packet: the STREAM frames for healthy requests coalesced into it, and the ACK it carried. The property your new test proves for one packet stops holding once the failed stream has burned its window, and the window is `initial_max_data`, 1 MiB by default (flare/quic/state.mojo:281), so a server already streaming a large body into a response we rejected at byte zero gets there in well under a second.

`cancel_stream` at flare/quic/client.mojo:1363 is the fix and it is sitting in this diff with no caller outside the new test. One STOP_SENDING when the reader first fails makes the peer stop and lets you drop the entry from `_pending` instead of carrying it for the life of the connection. #21 calls it from `H3Stream.close`, which covers the caller-abandons case, not this one.

This is the thing I want fixed in this PR rather than a follow-up, because the PR is specifically about isolating stream failures, and this is the path where the isolation gives out.

## 2. Two halves of RFC 9000 §3 that cancellation leaves open

`cancel_stream` sends the frames and records nothing locally, so `send_stream` on the same id afterwards still emits STREAM frames, which §3.1 forbids once RESET_STREAM has gone out. It is reachable from #21: `H3Stream.close` cancels, and an upload loop that is mid-`send_request_body` on the same connection does not know. Inline suggestion is to mark the stream RESET_SENT and make `send_stream` refuse it.

The receive half is missing too. `apply_stop_sending` (flare/quic/state.mojo:474-482) flips the stream to RESET_SENT and stops there, and nothing anywhere emits the RESET_STREAM that §3.5 requires in response, so a server that asks us to stop uploading gets silence and we keep sending. `QuicClientConnection.stream_reset` (flare/quic/client.mojo:1401-1406) tests only RESET_RECVD, so nothing above notices either - #23's `send_chunked` is where that starts to bite, since its upload loop keeps encoding DATA into a stream the peer has asked it to abandon. That one is outside this diff, so I am happy to take it as a follow-up if you would rather not grow this PR; say the word and I will open it.

## 3. Zero capacity waits 30 seconds and never says we are blocked

`_send_request_wire` spins on `send_capacity` and raises `h3 client: upload flow-control timeout` after 30 s of no progress (flare/http3/client.mojo:436). The bound itself is fine - the clock resets after every successful chunk, so a slow but live peer that grants credit at least once per 30 s is safe. What bothers me is that we never send STREAM_DATA_BLOCKED or DATA_BLOCKED. RFC 9000 §4.1 wants those exactly here, and without them a peer that grants credit on demand has nothing to react to. `derive_peer_send_limits` defaults a missing `initial_max_stream_data_bidi_remote` to 0 (flare/quic/transport_params.mojo:218-223), so against such a peer `send_capacity` is 0 from the start, we sit silent for 30 s, and the error we raise blames a timeout. On `main` the same case failed instantly with the limit in the message, which was at least diagnosable.

## 4. Tests

`test_stream_error_preserves_other_responses_in_same_packet` is the right test and it covers the two cases that matter most, reset and parse failure, in the same packet as a good response. The PTO test is nicely done, especially dropping the datagram and back-dating `time_ms` instead of sleeping.

The reader side is thin for the number of rules the diff adds. Three tests now stand behind: final-header ordering, duplicate and out-of-order `:status`, the 101 rejection, trailer field rejection, the connection/transfer-encoding/keep-alive/proxy-connection/upgrade set, header-name case, the field-size bound in two places, Content-Length parsing and conflict, body over Content-Length, body under Content-Length at FIN, DATA before HEADERS, DATA on a bodyless 204/304, bytes after FIN, control-stream frame types on a request stream, and unknown-frame skipping. Cases I would like before merge, since each is a few lines against the reader with no socket:

- Content-Length shorter and longer than the DATA total (the over case is the one that currently short-circuits mid-frame).
- 204 or 304 with a non-empty DATA frame.
- SETTINGS or PUSH_PROMISE on a request stream.
- A reserved or grease frame type with a large length, to show the skip is incremental and the inbox stays small.

Two more I am happy to see as follow-ups: MAX_STREAM_DATA lost and then recovered with the value having advanced in between (inline on the PTO test), and STOP_SENDING arriving from the server.

## 5. Smaller things, all inline

Every bodyless request now costs two ack-eliciting packets instead of one, because the FIN always rides its own zero-length STREAM frame. `release_stream_credit` emits an ack-eliciting datagram per stream per poll with no threshold, and it grows `conn.max_data_recv`, which doubles as the seed for every stream created later. The `_header_bytes` accumulator is never reset between field sections, so interim responses and trailers are charged against the same budget as the final head.

One that I could not anchor: `_apply_head` calls `_parse_status`, which raises rather than calling `_fail`, so a malformed `:status` throws out of `feed` with `_drain`'s cursor discarded and the inbox uncompacted. `poll_responses` catches it and converts it to a reader error, so the multiplexed path is fine, but `read_response` / `_feed_ext` has no such catch, and there the reader stays alive with the bad frame still at the head of its inbox and re-raises on every poll. Either make `_parse_status` non-raising through `_fail`, or wrap the `_feed_ext` push the way you wrapped the one in `poll_responses`.

## Commit message

The unique commit is a subject line with an empty body. The reasoning in the PR description is good and it belongs in the commit, since that is where this repo keeps rationale. Please move it over on the rebase.

## What I'll take

- The STOP_SENDING reply in `flare/quic/state.mojo`, if you would rather keep this PR to the client.
- The MAX_DATA update threshold and the retransmit-the-current-value question, as a loss-recovery follow-up with the rest of the RFC 9002 gaps already listed in the module docstring.

From you: item 1, item 2's send-after-reset half, the blocked frames in item 3, the four reader tests, and the commit body. The rest can follow.

ehsanmok:
Landed on \`main\` as 39bfee4, cherry-picked clean with your authorship intact.

Two follow-ups, both mine:

- cdb59f2 regenerates \`tests/_agg\` for the response-reader tests.
- df7b858 closes the review items. A failed reader now sends STOP_SENDING once via \`cancel_stream\` on the first transition into the error state -- releasing zero credit was right for the bytes being discarded, but on its own it stranded the connection: \`max_recv_data\` stopped moving while \`recv_offset\` kept climbing, and past the limit the client drops whole packets, healthy streams and ACKs included. With the default 1 MiB that arrives quickly, so the isolation this PR is about stopped holding exactly when it was needed. \`cancel_stream\` also marks the stream RESET_SENT before emitting, so a later \`send_stream\` cannot emit a STREAM frame (RFC 9000 sec 3.1) and a PTO retransmit cannot carry a final size the peer never saw. And the response reader resets its field-section counter on a 1xx, since SETTINGS_MAX_FIELD_SECTION_SIZE is per-section.

Three new tests; the per-section budget and the reset contract each fail before.

The remaining items from that review -- inbound STOP_SENDING producing a RESET_STREAM, and the MAX_DATA update threshold -- are still open and not in this. Thank you. Closing since the commit is on main.

PR #21, opened 2026-09-04
feat(http): add cross-protocol response readers

bowyern:
## Summary

- add exclusive-connection HTTP/2 and HTTP/3 download adapters behind a protocol-erased `HttpStreamResponse`
- provide bounded pulls, limited collection, trailers at end of stream, and idempotent early cancellation
- normalize request headers once across HTTP versions, stripping caller framing and connection-nominated fields and rejecting invalid names or CR/LF values
- implement bounded h2c upgrade response handling for the streaming path

## Why

This internal layer gives all three protocols the same ownership and cleanup contract before the API is exposed on `HttpClient`. Central header preparation also prevents protocol-specific differences from reintroducing request smuggling or header-injection behavior.

## Stack

Part 5 of 8. **Stacked on #20; review #17–#20 first.**

The GitHub diff is cumulative against `main` until the preceding PRs merge. The unique commit for this layer is `616e03b`.

## Verification

- `tests/http/test_stream_request_headers.mojo`


flare/http/_client/stream_request.mojo
```diff
@@ -0,0 +1,190 @@
...
+            headers.set("TE", "trailers")
+    for i in range(headers.len()):
+        var key = headers._keys[i]
+        var value = headers._values[i]
+        if (
+            not _valid_token(key)
+            or value.find("\r") >= 0
+            or value.find("\n") >= 0
+        ):
+            raise Error("HTTP streaming: invalid request header")
```
ehsanmok, line 87: This is the CR/LF half of the check. NUL, the other C0 controls, DEL and a leading or trailing SP/HTAB all get through, and on h2/h3 every one of those makes the request malformed at the receiver (RFC 9113 §8.2.1, RFC 9114 §4.2). `stream_qpack_headers` at :130 and `_build_h2_request_headers` at `flare/http/_client/h2_send.mojo:49` hand the value straight to the encoder, so this loop is the last place to catch it. CR and LF stay covered below since both are under 32.

```suggestion
    for i in range(headers.len()):
        var key = headers._keys[i]
        var value = headers._values[i]
        if not _valid_token(key):
            raise Error("HTTP streaming: invalid request header")
        var bytes = value.as_bytes()
        for b in bytes:
            if (b < 32 and b != 9) or b == 127:
                raise Error("HTTP streaming: invalid request header")
        if len(bytes) > 0 and (
            bytes[0] == 32
            or bytes[0] == 9
            or bytes[len(bytes) - 1] == 32
            or bytes[len(bytes) - 1] == 9
        ):
            raise Error("HTTP streaming: invalid request header")
```

flare/http/_client/stream_request.mojo
```diff
@@ -0,0 +1,190 @@
...
+    headers: HeaderMap,
+    upload: Bool = False,
+    upgrade: Bool = False,
+) raises -> String:
+    if not _valid_token(method) or method.upper() == "CONNECT":
+        raise Error("HTTP streaming: unsupported request method")
+    var target = u.request_target()
+    for b in target.as_bytes():
+        if b <= 32 or b == 127:
+            raise Error("HTTP streaming: invalid request target")
```
ehsanmok, line 103: The target gets a byte-by-byte check and then the authority goes into the `Host:` line five lines down untouched. `Url.parse` does not check host characters either, `flare/http/url.mojo:193` only rejects an empty host, so a host carrying a bare LF lands in the request line. DNS kills that before it reaches a socket on the direct path, but `_connect_tunnel` (`flare/http/client.mojo:1286`) hands the same unvalidated host to a proxy inside a `CONNECT` line. Cheap to close here, in the function that is already doing the validating. Worth reusing `authority` on line 109 rather than calling `_h2_authority` twice.

```suggestion
    var target = u.request_target()
    for b in target.as_bytes():
        if b <= 32 or b == 127:
            raise Error("HTTP streaming: invalid request target")
    var authority = _h2_authority(u)
    for b in authority.as_bytes():
        if b <= 32 or b == 127:
            raise Error("HTTP streaming: invalid request authority")
```

flare/http/_client/stream_response.mojo
```diff
@@ -0,0 +1,152 @@
...
+            return
+        if self._kind == 1:
+            Pool[HttpDownload[_H2Transport]].free(self._addr)
+        elif self._kind == 2:
+            Pool[Http2Download].get_ptr(self._addr)[].close()
+            Pool[Http2Download].free(self._addr)
+        else:
+            Pool[Http3Download].get_ptr(self._addr)[].close()
+            Pool[Http3Download].free(self._addr)
+        self._addr = 0
```
ehsanmok, line 152: `__deinit__` at :53 calls this and the docstring promises it survives an error, so it has to be total. The frees run before `_addr` is cleared, so a raise out of `Http2Download.close()` leaks the fd and leaves `_addr` set for the next call to retry the same cancel. Clearing the flag first makes the state transition atomic whatever the driver does.

```suggestion
    def close(mut self):
        """Release the connection. Safe after EOS, an error, or another close.
        """
        if self._addr == 0:
            return
        var addr = self._addr
        self._addr = 0
        if self._kind == 1:
            Pool[HttpDownload[_H2Transport]].free(addr)
        elif self._kind == 2:
            Pool[Http2Download].get_ptr(addr)[].close()
            Pool[Http2Download].free(addr)
        else:
            Pool[Http3Download].get_ptr(addr)[].close()
            Pool[Http3Download].free(addr)
```

flare/http/_client/h2_stream.mojo
```diff
@@ -0,0 +1,110 @@
...
+
+    def close(mut self):
+        if not self._done:
+            self.conn.cancel_stream(self.sid)
+            try:
+                self.flush()
+            except:
+                pass
+        self._done = True
+        self.transport.close()
```
ehsanmok, line 110: `flush()` is guarded and `cancel_stream` is not, though both can raise. `HttpStreamResponse.close()` calls this at `flare/http/_client/stream_response.mojo:147` and only frees the pool cell on the line after, so a raise here skips the free. The h3 driver already wraps its whole close body at `h3_stream.mojo:82-87`; same shape.

```suggestion
    def close(mut self):
        if not self._done:
            try:
                self.conn.cancel_stream(self.sid)
                self.flush()
            except:
                pass
        self._done = True
        self.transport.close()
```

flare/http/_client/h3_stream.mojo
```diff
@@ -0,0 +1,88 @@
...
+        self.sid = sid
+        self.status = 0
+        self.headers = HeaderMap()
+        self.trailers = HeaderMap()
+        self._pending = List[UInt8]()
+        self._pos = 0
+        self._done = False
+        self._timeout_ms = timeout_ms
+        var start = _monotonic_ms()
+        while not self.conn.head_ready(sid):
```
ehsanmok, line 33: No error check in this loop. The h2 constructor calls `_check_error()` on every pass at `h2_stream.mojo:37-40` and leaves on GOAWAY or a stream reset; here a reset stream burns the whole timeout and then reports a timeout, which sends the caller looking in the wrong place.

The bound is the part I want fixed. `_check_timeout` is a no-op when `_timeout_ms <= 0` (:45), and `HttpClient(timeout_ms=0)` is legal, since `flare/http/client.mojo:313` only defaults it to 30_000. On that client a server that opens the stream and never sends a head puts this in an endless 100 ms poll with nothing to break it. A hard cap would do for now, ahead of any stream-error API to check against.

tests/http/test_stream_request_headers.mojo
```diff
@@ -0,0 +1,27 @@
...
+    headers.append("Authorization", "first")
+    headers.append("Authorization", "second")
+    headers.set("Accept-Encoding", "gzip")
+    var prepared = prepare_stream_headers(headers, 3, "agent", "Bearer stored")
+    assert_equal(prepared.get("X-Hop"), "")
+    assert_equal(prepared.get("Content-Length"), "3")
+    assert_equal(len(prepared.get_all("Content-Length")), 1)
+    assert_equal(len(prepared.get_all("Authorization")), 1)
+    assert_equal(prepared.get("Authorization"), "Bearer stored")
+    assert_equal(prepared.get("Accept-Encoding"), "gzip")
```
ehsanmok, line 23: The defaults this function injects are worth pinning as well, since `_stream_exchange` in #22 leans on them being present exactly once.

```suggestion
    assert_equal(prepared.get("Accept-Encoding"), "gzip")
    assert_equal(prepared.get("User-Agent"), "agent")
    assert_equal(prepared.get("Accept"), "*/*")
```

The rejection cases are the ones I actually need here; they are listed in item 2 of the review.

flare/http/_client/stream_response.mojo
```diff
@@ -0,0 +1,152 @@
...
+                    self.trailers = p[].trailers.copy()
+            if ended:
+                self.close()
+            return out^
+        except e:
+            self.close()
+            raise e^
+
+    def read_all(mut self, max_bytes: Int = 65536) raises -> List[UInt8]:
+        """Buffer the remaining body. ``max_bytes`` is the pull size, NOT a limit.
```
ehsanmok, line 110: No bound on this path. It becomes reachable in #23, where `send_chunked` calls it on a response from an arbitrary server (`flare/http/client.mojo:1142` on that branch) before handing back a buffered `Response`, so the whole body lands in memory with nothing to stop it. `read_all_limited` is right below. I'm recording it here rather than on #23 because this is where the default lives, and the buffered `send()` path has the same gap, so the follow-up is mine either way.

While you are in the docstring, the shouty `NOT` can go; the sentence carries the point on its own.

ehsanmok (changes requested):
Thanks for this - the protocol-erased `HttpStreamResponse` is the right abstraction, and the ownership contract behind it comes out the same on all three wires, which is what I most wanted to check. I read the layer commit against the tree and ran what there is to run before writing this. Four things before merge. The summary calls `prepare_stream_headers` the thing that stops per-protocol differences from reintroducing smuggling and header injection, and it checks CR and LF and nothing else, with the authority not checked at all. Both `close` paths do work that can raise before they release the pool cell, on an object whose docstring promises `close` is safe after an error. Nothing in the PR exercises any of the rejection paths the layer adds. And an h3 stream on a client configured without a timeout polls forever. The rest below I'll take myself.

On the rebase: this conflicts with `main` only in `flare/tls/stream.mojo`, inherited from #17. Run `pixi run tests-gen` after rebasing so `tests/http/test_stream_request_headers.mojo` joins `tests/_agg`; see my note on #17 for what that runner does. A clean textual merge is not a clean compile after the API migration, so I'd like to watch this layer build on current `main` as well.

## What I ran

Worktree at `616e03b`, Mojo 1.0.0 (ed45d567), macOS arm64.

| entry | result | wall |
|---|---|---|
| `tests/http/test_stream_request_headers.mojo` | 1 test run: 1 passed, 0 failed, 0 skipped | 1 s |
| `tests/http/test_client_stream_download.mojo` | 3 passed | 5 s |

`mojo format` is clean on the 22 changed `.mojo` files. One test function against four new files is the number behind items 2 and 3 below. Everything else here comes from reading the code, not from a run: no test in this PR, #22 or #23 ever constructs an `Http3Download`, so the h3 reader is uncovered end to end.

## 1. The header check that is meant to stop injection looks only for CR and LF

`flare/http/_client/stream_request.mojo:82-87` rejects a value containing `\r` or `\n`, and a name that fails `_valid_token`. Everything else goes through: NUL, the rest of the C0 controls, DEL, and a leading or trailing SP or HTAB. On h2 and h3 each of those makes the request malformed at the receiver (RFC 9113 §8.2.1, RFC 9114 §4.2), and this loop is the last place to catch them, since `stream_qpack_headers` at `:130` and `_build_h2_request_headers` at `flare/http/_client/h2_send.mojo:49-70` hand the value straight to the encoder.

Where it does hold: `HeaderMap.set` and `.append` already raise `HeaderInjectionError` on CR/LF, through `_check_injection` at `flare/http/headers.mojo:76-86` (called at `:149` and `:168`), so for a caller going through those this loop is a second line of defence. `set_unchecked` does not, which is what makes the loop load-bearing rather than decorative. Suggestion inline.

The authority has no check at all. `Url.parse` only rejects an empty host (`flare/http/url.mojo:194-195`), and `stream_h1_head` drops `_h2_authority(u)` into the `Host:` line at `:109` untouched, five lines after giving the request target a byte-by-byte scan. DNS kills a bad host before it reaches a socket on the direct path, but `_connect_tunnel` (`flare/http/client.mojo:1286`) hands the same unvalidated host to a proxy inside a `CONNECT` request line. That is the half I'd rather not leave for later, and it is four lines in the function that is already doing the validating. Suggestion inline.

## 2. Both `close` paths release the slot after work that can raise

`HttpStreamResponse.close` at `flare/http/_client/stream_response.mojo:139-152` is called from `__deinit__` at `:53`, and its docstring says it is safe after EOS, after an error, and after another close. It is not total as written. The frees run before `self._addr = 0`, so a raise out of `Http2Download.close()` leaks the fd and leaves `_addr` set, and the next call retries the same cancel on a half-torn-down object. Clearing the flag first makes the transition atomic whatever the driver does.

`H2Stream.close` at `flare/http/_client/h2_stream.mojo:102-110` is the driver in question. `flush()` is wrapped, `cancel_stream` is not, and both can raise. The h3 driver already wraps its whole close body at `h3_stream.mojo:82-87`, so this is one shape being applied in two places out of three. Suggestions inline on both.

Neither case is exotic. An h2 cancel writes RST_STREAM and flushes it, and the peer having gone away is both what makes that write raise and what makes `close` run in the first place.

## 3. Nothing exercises a rejection

One test, and it asserts the strip list and the injected defaults on a well-formed input. Not one of the raise paths this layer adds has a case: the invalid-name and CR/LF rejections at `stream_request.mojo:87`, `CONNECT` and a non-token method at `:98-99`, the request-target scan at `:100-103`, the `Expect` refusal at `:59-61`, the h2c bound at `:155` and `:163`, and the non-101 fallback at `:176-179`. #22 and #23 sit on top of this and add none of them either, so the whole family is untested across three PRs.

These need no socket. `prepare_stream_headers` and `stream_h1_head` are ordinary functions over a `HeaderMap` and a `Url`, and the existing test already shows the shape. A case per raise is the ask. It is also what keeps item 1's fix from quietly coming undone the next time this function is edited.

## 4. An h3 stream with no deadline polls forever

`flare/http/_client/h3_stream.mojo:33` loops on `head_ready` with no error check and no hard bound. That is two problems in the same loop.

`_check_timeout` is a no-op when `_timeout_ms <= 0` (`:45`), and `HttpClient(timeout_ms=0)` is legal - `flare/http/client.mojo:313` only defaults the field to 30000. On such a client, a server that opens the stream and never sends a head parks this in a 100 ms poll with nothing to break it. The h2 constructor calls `_check_error()` on every pass (`h2_stream.mojo:37-40`) and leaves on GOAWAY or a stream reset; here a reset stream burns the whole timeout and then reports a timeout, which sends the caller looking in the wrong place.

A hard cap would do for now, ahead of any stream-error API to check against on the h3 side.

Listing these so nobody re-argues them:

- Framing and hop-by-hop stripping is casing-proof. `HeaderMap.remove` (`:215`), `contains` (`:201`) and `get` (`:172`) compare case-insensitively (`flare/http/headers.mojo:92`), so the strip list at `stream_request.mojo:48-58` catches `content-length` and `Content-Length` alike.
- `Host` is always regenerated, never forwarded. `stream_h1_head:108-109` builds it from the URL, and the h2 and h3 paths go through `_h2_authority(u)`. A caller-supplied `Host` is dropped on all three wires.
- Name lowercasing is done on both binary wires: `_build_h2_request_headers` (`h2_send.mojo:49-70`) and `stream_qpack_headers` (`:130-137`).
- `Pool.alloc_move` never returns 0, so the `_addr == 0` sentinel in `HttpStreamResponse` is sound rather than lucky.
- `_H2Transport.__deinit__` and `.close()` are idempotent and free both pooled streams, so the kind-1 arm of `HttpStreamResponse.close()` does release the socket through the download's destructor. No arm returns anything to a pool, which is the exclusivity claim in the summary.
- The `_collect` bound at `stream_response.mojo:123-137` pulls limit+1 to detect the overshoot, so a body of exactly the limit passes and `n+1` raises. That is the behaviour the docstring claims.
- The `status < 200` guard at `h2_stream.mojo:47-48` is unreachable: `flare/http2/state.mojo:768-772` returns early on a 1xx without setting `headers_complete`, so 103 Early Hints never reaches it. Harmless, and I'd delete it.
- The h2c upgrade bound is 65536 across the informational heads plus the final head (`stream_request.mojo:155`, `:163`), a non-101 falls back to h1 with the bytes already read (`:176-179`), and leftover bytes after a 101 are fed in as h2 frames, which is right per RFC 7540 §3.2.
- `Content-Length` handling lines up with the layers above: #22 passes `len(body)` and #23 passes -1 with `Transfer-Encoding: chunked` added by `stream_h1_head:114-115`.

## What I could not verify

Whether `punctuation.find(chr(0))` inside `_valid_token` returns -1 or 0, so whether a NUL in a header *name* is rejected today. That needs a scratch program against the compiler and I didn't write one; it does not change item 1, which is about values. Whether `Http2Download.close()` raises in practice: `_send_rst_stream` does no raising work, but both it and `cancel_stream` are `def`s, so the path exists at the type level and `close` has to be written for it. And anything about this branch rebased, since it has not been.

The suggestion blocks below were checked against the names in scope and against equivalent idioms in the tree (`flare/http2/state.mojo:748-757` for the `as_bytes()` loop, `h3_stream.mojo:82-87` for the try/except shape). They were not compiled.

## What I'll take

- End-to-end h3 coverage of `Http3Download`, including a wire assertion that a cancel goes out exactly once. It needs a live h3 fixture, so it is not yours to build.
- The unreachable `status < 200` guard at `h2_stream.mojo:47`.
- The same unvalidated host in `_connect_tunnel` on `main`, if you'd rather not widen this PR past the streaming path.
- The unbounded `read_all` default, inline on `stream_response.mojo:110`.

Dropping a `TE` list like `trailers, deflate` wholesale rather than reducing it to `trailers` is conservative and I'm happy to leave it.

From you: items 1 and 2, the rejection tests in item 3, and the bound in item 4. The commit message for this layer is a subject line with nothing under it, and the "Why" paragraph in the description is exactly what this repo keeps in a commit body, so please move it across on the respin.

ehsanmok:

here.

PR #22, opened 2026-09-04
feat(http): expose streamed responses through HttpClient

bowyern:
## Summary

- expose `send_streaming`, `get_streaming`, and the compatibility `get_streaming_tls` wrapper on `HttpClient`
- dispatch streamed responses across HTTP/1.1, h2 prior knowledge, h2c upgrade, TLS ALPN, HTTP/3, and configured proxies
- apply retries and redirects only before returning the response, strip credentials on cross-origin redirects, and never replay body reads
- keep streamed connections exclusive and compressed response bodies raw

## Why

This is the public download API. Its policy boundaries are explicit because retrying after bytes reach the caller, pooling a partially consumed connection, or forwarding authorization across origins would be unsafe.

## Stack

Part 6 of 8. **Stacked on #21; review #17–#21 first.**

The GitHub diff is cumulative against `main` until the preceding PRs merge. The unique commit for this layer is `ddfbb35`.

## Verification

- `tests/http/test_client_stream_response.mojo`
- `tests/http/test_client_stream_download_tls.mojo`
- `tests/http/test_client_proxy.mojo`
- `tests/http/test_h3_live_dial.mojo`


ehsanmok (commented):
Thanks for this - it is the layer the rest of the stack was building toward, and it does what issue #5 actually asked for: `get_streaming` takes `https://` now and picks the wire instead of pinning ALPN to `http/1.1` (the ALPN half of that lands in #17, where I raised it). I reproduced all four test entries locally at `ddfbb35` before reviewing and they are green. The boundary I cared about most, that no retry can fire once a body byte has reached the caller, holds because of how the code is laid out rather than by convention, and I traced it. What I need before merge is a rebase I want to watch compile, and one claim in the description narrowed. The rest I am happy to take myself.

On the rebase: `flare/tls/stream.mojo` and `tests/http/test_client_proxy.mojo` conflict with `main`. Run `pixi run tests-gen` after rebasing so `tests/http/test_client_stream_response.mojo` joins `tests/_agg`; see my note on #17 for what that runner does. One extra piece here that shows up as no conflict at all: `tests/_agg/agg_http.mojo:385` imports `test_get_streaming_rejects_https`, the symbol this PR renames, and registers it again at `:3812`, so the tree compiles locally and `pixi run tests` breaks.

## What I verified

| entry | result | wall |
|---|---|---|
| `tests/http/test_client_stream_response.mojo` | 9 tests run, 9 passed | 24s |
| `tests/http/test_client_stream_download_tls.mojo` | 3 tests run, 3 passed | 22s |
| `tests/http/test_client_proxy.mojo` | 2 passed | 18s |
| `tests/http/test_h3_live_dial.mojo` | 4 passed | 24s |

`mojo format` clean on the 28 changed files. Every build in those logs also prints `./flare/http/_client/stream_request.mojo:150:15: warning: assignment to 'end' was never used`. That file arrives in #21 so it is #21's to drop, but it is noisy on every compile of this layer.

Read rather than run:

- The retry boundary is structural. `send_streaming` (`flare/http/client.mojo:1396-1408`) wraps `_stream_once`, and `_stream_once` returns the moment the head is parsed, so every retry and every redirect hop finishes before the caller can reach `read_chunk`. A 5xx retry closes the previous response at `client.mojo:1401` and re-dials. Nothing leads from a body read back into that loop.
- Exclusivity holds on all three wires. `_stream_transport` (`client.mojo:1479-1501`) always dials fresh and never touches `self._pool` or `self._tls_pool`, and `_dial_http3` (`client.mojo:793-828`) dials fresh too, reporting the miss to `_quic_pool` rather than acquiring from it. `HttpStreamResponse.close` frees its slot instead of releasing it, so even a fully drained connection is dropped rather than pooled. On h1 the `Connection: close` at `flare/http/_client/stream_request.mojo:126` forces that anyway; on h2 and h3 it leaves reuse on the table, which I am content to pick up as a follow-up.
- The compressed-body question answers itself. `prepare_stream_headers` sets `Accept-Encoding: identity` when the caller did not set one (`stream_request.mojo:67-68`), so a default streamed GET never asks for gzip and only a caller who sets the header gets raw compressed bytes back. `test_client_stream_response.mojo` covers that case with a real gzip body. It diverges from the buffered path, which sends `gzip, deflate, br` at `client.mojo:1873`, and the divergence is the right call for a reader that cannot buffer.
- Cross-origin is scheme plus host plus port, and `Url.parse` fills in 80 / 443 (`flare/http/url.mojo:169-190`), so `http://h` and `http://h:80` compare equal. Host is compared byte for byte, so `Example.com` and `example.com` read as two origins and get more stripping rather than less, which is the safe direction to be wrong in.
- 307 / 308 has an easy answer here. `send_streaming` takes a buffered `Request`, so replaying its body across a hop costs nothing and the code does replay it; 303 drops the body and the `Content-Type` with it at `client.mojo:1457-1459`, and the hop cap surfaces as `TooManyRedirects` at `client.mojo:1477`.
- The deleted lines in `tests/http/test_client_stream_download_tls.mojo` are docstring prose plus the old test body. No assertion was lost; the renamed test asserts the same two things about the opposite direction.

## 1. Narrow the credential claim

The summary says this strips credentials on cross-origin redirects. It strips one credential: `Authorization`, at `client.mojo:1460-1462`, and `test_cross_origin_redirect_strips_auth` proves it across two ports. The cookie jar is re-attached at the top of every hop (`client.mojo:1433-1435`), and that jar is origin-agnostic on purpose, documented at `flare/http/_client/cookie_store.mojo:15-19`, so a session cookie set by origin A rides along to origin B. A caller-supplied `Cookie` header survives too, since only `Authorization` is removed from the map.

The buffered path has the identical hole at `client.mojo:1876-1878`, so it is not a regression and I am not asking you to scope the jar in this PR. I do want the bullet to say `Authorization` instead of credentials, so nobody reads the summary and assumes cookies are handled. Per-origin jar scoping is mine.

## 2. Docstrings

The new entry points lost their docstrings down to one line each. `get_streaming` used to explain why there were two of them; it now reads "GET a streamed response over the selected HTTP protocol" and says nothing about the behavior change that is the point of the PR. Someone reading it cannot tell that https is accepted, which wire gets picked, or that the connection is theirs alone. Suggestion inline, and a second one for the `Accept-Encoding: identity` default on `send_streaming`, which is the one behavior here likely to catch a caller out.

## 3. Commit message

The unique commit for this layer is a bare subject line with an empty body. The "Why" paragraph in the PR description is exactly what this repo keeps in commit messages, so please move it across. The policy boundaries and the reason each one exists are what someone will want when they `git log` this in a year, and the PR description is not where they will look.

## 4. Test gaps, none of them blocking

Nine tests is decent coverage for a surface this wide. What is missing, in the order I would add it:

- TLS through a proxy. `test_client_proxy` runs both buffered and streaming now, but cleartext only. CONNECT then TLS then ALPN in `_stream_transport` is the path with the most moving parts and has no test on either side.
- h3 with a proxy configured. `client.mojo:1519` skips the h3 branch whenever a proxy is set and falls through to the tunnel, which is the right behavior, and nothing pins it.
- The https happy path through `get_streaming_tls`. `_drain` moved to `get_streaming`, so the compatibility wrapper is now only tested for rejecting cleartext. Inline.
- `Authorization` surviving a same-origin redirect. The `/redirect` case asserts the cookie arrives but never checks the auth header, so half the policy is pinned.
- An h2c upgrade the server declines. `finish_stream_h1` has a branch that hands the buffered non-101 head to `HttpDownload` (`stream_request.mojo:176-179`) and it is unexercised.

I will take any of these as follow-ups rather than hold the stack for them.

## What I'll take

Per-origin cookie jar scoping, h2 and h3 connection reuse on the streaming path, and any of the five test gaps in item 4 as follow-ups rather than holding the stack for them.

From you: the rebase, with `tests-gen` run and the `agg_http.mojo` import fixed, so I can watch it compile against current `main`; the narrowed credential bullet; the commit body; and the docstrings if you have the patience for them, though say so if not and I'll do them on merge.

`docs/features.md:95` and `docs/cookbook.md:118` still describe the old two-entry-point `-> HttpDownload` API, and that is already fixed in #24 later in the stack, so nothing to do here.

flare/http/client.mojo
```diff
@@ -1350,103 +1366,206 @@ struct HttpClient(Movable):
...
-        """GET a response and stream its body incrementally.
-
-        Parses only the status line + headers up front and returns an
-        :class:`HttpDownload`; the caller pulls the body with
-        ``read_chunk()`` in bounded memory (Content-Length, chunked, and
-        close-delimited framings decoded on the fly). The connection is
-        not pooled (``Connection: close``).
+    def get_streaming(self, url: String) raises -> HttpStreamResponse:
+        """GET a streamed response over the selected HTTP protocol."""
+        return self.send_streaming(Request(method="GET", url=url))
```
ehsanmok, line 1371: This is the entry point the PR exists for, and the docstring is the one place that does not say so. On `main` this raised on `https://`; now it accepts it and selects the wire. One line cannot carry that, nor the exclusivity contract that makes the whole design safe. House rule here is that a docstring describes what the code does now, so no "used to" wording. This is what I checked against the branch:

```suggestion
    def get_streaming(self, url: String) raises -> HttpStreamResponse:
        """GET a streamed response over ``http://`` or ``https://``.

        The wire follows the client configuration: HTTP/3 when the
        Alt-Svc / http3 policy selects it and no proxy is configured,
        otherwise ALPN ``h2`` or ``http/1.1`` over TLS, and h2c or
        HTTP/1.1 in cleartext. The response owns its connection
        exclusively; it is never taken from or returned to a pool.

        Args:
            url: The target URL (absolute or relative to ``base_url``).

        Returns:
            An :class:`HttpStreamResponse` positioned at the first body
            byte.

        Raises:
            NetworkError: On connection, TLS or I/O failure.
            TooManyRedirects: If the redirect limit is exceeded.
        """
        return self.send_streaming(Request(method="GET", url=url))
```

flare/http/client.mojo
```diff
@@ -1350,103 +1366,206 @@ struct HttpClient(Movable):
...
-    def _streaming_get_wire(self, u: Url, default_port: Int) -> String:
-        """Build the HTTP/1.1 GET head shared by both streaming paths.
+    def send_streaming(self, req: Request) raises -> HttpStreamResponse:
+        """Send a buffered request; return its final head before reading its body.
 
-        ``Accept-Encoding: identity`` on purpose: a streaming download
-        pulls bounded chunks, and a compressed body would have to be
-        fully buffered to inflate, which is the opposite of the point.
+        Uses ordinary redirect/retry policy only until this method returns.
+        Body reads are never replayed. Content encodings are not decompressed.
```
ehsanmok, line 1386: "Content encodings are not decompressed" reads as though any body might come back gzipped, which is the scarier half of the truth. `prepare_stream_headers` puts `Accept-Encoding: identity` on the request when the caller did not set one (`flare/http/_client/stream_request.mojo:67-68`), so a plain `get_streaming` never asks for a coding in the first place. Only a caller who sets the header opts into raw bytes. Worth saying, because the buffered path does the opposite at `client.mojo:1873`.

```suggestion
        Body reads are never replayed. ``Accept-Encoding`` defaults to
        ``identity``; a caller who asks for a content coding gets the
        encoded bytes back undecoded.
```

flare/http/client.mojo
```diff
@@ -1350,103 +1366,206 @@ struct HttpClient(Movable):
...
+            if decision.action == RedirectAction.FOLLOW:
+                response.close()
+                url = decision.next_url
+                method = decision.next_method
+                if decision.next_body_dropped:
+                    body = List[UInt8]()
+                    _ = headers.remove("Content-Type")
+                if not decision.forward_authorization:
+                    auth = ""
+                    _ = headers.remove("Authorization")
```
ehsanmok, line 1462: This is the only credential the cross-origin hop drops, and the PR summary says credentials. Two others survive it.

The jar is re-attached at the top of the next iteration at `client.mojo:1433-1435`, and it is origin-agnostic by design - `flare/http/_client/cookie_store.mojo:15-19` spells that out - so a session cookie set by origin A goes to origin B. A `Cookie` header the caller set by hand survives as well, since `headers.remove` here only names `Authorization`. requests drops both on a cross-host redirect.

The buffered path does exactly the same thing at `client.mojo:1876-1878`, so I am not treating it as a regression and I am not asking you to scope the jar inside this PR. Please just narrow the summary bullet to `Authorization`. The per-origin jar is on my list.

flare/http/client.mojo
```diff
@@ -1350,103 +1366,206 @@ struct HttpClient(Movable):
...
-        var proxy = self._resolve_proxy(u)
+            var response = self._stream_exchange(method, url, prepared, body)
+            self._record_stream_head(url, response.headers)
+            if (
+                response.status != 301
+                and response.status != 302
+                and response.status != 303
+                and response.status != 307
+                and response.status != 308
+            ):
```
ehsanmok, line 1447: The buffered path gates on `Response.is_redirect()` at `client.mojo:1903`, which is the whole 300-399 range. This explicit set of five means a 300 or a 305 carrying a `Location` is now handed back to the caller instead of being followed. I think the explicit set is the better rule and I would keep it, but the two paths disagreeing silently is the sort of thing that bites a year from now. A line in the commit body saying which codes are followed, or an assertion on a 300, would settle it.

flare/http/client.mojo
```diff
@@ -1350,103 +1366,206 @@ struct HttpClient(Movable):
...
+            u,
+            headers,
+            upgrade=(
+                self._h2c_upgrade and not u.is_tls() and not self._prefer_h2c
+            ),
+        )
+        var proxy = self._resolve_proxy(u)
+        if (
+            u.is_tls()
+            and proxy.byte_length() == 0
```
ehsanmok, line 1519: Good call: with a proxy configured the h3 branch is skipped outright and the request goes down the CONNECT tunnel, rather than attempting QUIC through a proxy that cannot carry it. That is the behavior I would want, and it is quiet enough that nothing would notice if it changed. A test with `with_proxy` set against an `https://` origin, asserting `protocol() == "h2"` or `"http/1.1"`, would pin it. Follow-up is fine.

flare/http/client.mojo
```diff
@@ -1350,103 +1366,206 @@ struct HttpClient(Movable):
...
+            var s = conn.conn.streams[sid].copy()
+            s.response_body_allowed = method != "HEAD"
+            conn.conn.streams[sid] = s^
+            return HttpStreamResponse(Http2Download(transport^, conn^, sid))
+        transport.write_all(Span(wire.as_bytes()))
+        if len(body) > 0:
+            transport.write_all(Span(body))
+        return finish_stream_h1(
+            transport^, method, self._h2c_upgrade and not u.is_tls()
+        )
```
ehsanmok, line 1568: The same predicate is spelled two ways: line 1512 adds `and not self._prefer_h2c`, this one does not. I worked through the combinations and the mismatch is unreachable today, because a cleartext `prefer_h2c` client sets `transport._is_h2` at `client.mojo:1500` and never reaches here. It only holds by accident of another branch, though, so I would rather the two spellings match.

```suggestion
        return finish_stream_h1(
            transport^,
            method,
            self._h2c_upgrade and not u.is_tls() and not self._prefer_h2c,
        )
```

tests/http/test_client_stream_download_tls.mojo
```diff
@@ -75,7 +70,7 @@ def _drain(port: UInt16, path: String) raises -> Tuple[Int, Int]:
     """
     var url = String("https://localhost:") + String(Int(port)) + path
     with HttpClient(TlsConfig(ca_bundle=_CA_CRT)) as c:
-        var dl = c.get_streaming_tls(url)
+        var dl = c.get_streaming(url)
```
ehsanmok, line 73: Moving `_drain` onto `get_streaming` is right for what this file is testing now, but it takes the last https call to `get_streaming_tls` with it. The wrapper keeps its own `_resolve_url` and scheme check at `flare/http/client.mojo:1375-1380`, and after this the only thing exercising it is the cleartext rejection below. One drain through the wrapper, or a second `_drain` parameterised on which entry point it calls, gets the coverage back for a couple of lines.

tests/http/test_client_stream_download_tls.mojo
```diff
@@ -156,19 +151,19 @@ def test_https_streaming_download_chunked() raises:
     assert_true(largest <= _PULL_CAP, "pull exceeded the cap")
 
 
-def test_get_streaming_rejects_https() raises:
-    """The cleartext entry point still refuses https, with a pointer."""
+def test_get_streaming_tls_rejects_cleartext() raises:
```
ehsanmok, line 154: Rebase note rather than a review point: `main` picked up the per-area aggregate runner since you branched, and `tests/_agg/agg_http.mojo:385` imports this function under its old name `test_get_streaming_rejects_https`, registering it again at line 3812. Both need updating with the rename, and `tests/http/test_client_stream_response.mojo` needs `pixi run tests-gen` so it joins the aggregate at all. Neither shows as a merge conflict, so it will compile locally and break `pixi run tests`.

ehsanmok:

here.

PR #23, opened 2026-09-04
feat(http): stream request bodies across HTTP versions

bowyern:
## Summary

- add `send_chunked` for unknown-length `ChunkSource` request bodies over HTTP/1.1, HTTP/2, and HTTP/3
- generate protocol framing in the selected wire driver instead of trusting caller-supplied framing headers
- stop consuming the source when HTTP/2 returns an early response, including while upload flow control is exhausted
- make streamed uploads one-shot: no retries, redirects, pooling, or concurrent response-body exposure

## Why

An application-provided stream cannot generally be rewound safely. Keeping this path one-shot avoids duplicate side effects and partial-body replay, while early-response handling prevents a rejected upload from hanging behind withheld flow-control credit.

## Stack

Part 7 of 8. **Stacked on #22; review #17–#22 first.**

The GitHub diff is cumulative against `main` until the preceding PRs merge. The unique commit for this layer is `02d1fd3`.

## Verification

- `tests/http/test_client_stream_upload_protocols.mojo`
- `tests/http/test_h3_live_dial.mojo`


flare/http/client.mojo
```diff
@@ -1096,119 +1108,116 @@ struct HttpClient(Movable):
...
+                while True:
+                    var next_chunk = source.next(cancel)
+                    if not next_chunk:
+                        break
+                    var chunk = next_chunk.take()
+                    if len(chunk) > 0:
+                        h3.send_request_body(
+                            sid, chunk, False, self._timeout_ms
+                        )
+                h3.send_request_body(sid, List[UInt8](), True, self._timeout_ms)
```
ehsanmok, line 1197: This loop is the h3 counterpart of `upload_h2`, and it is missing that function's early-response check. `flare/http/_client/stream_request.mojo:239-241` returns as soon as a response is ready; here we pull the source dry first and only then construct `Http3Download`.

What makes it more than a missed optimization is the credit wait from #20. `_send_request_wire` (`flare/http3/client.mojo:423-443`) calls `poll_responses(100)` while `send_capacity(sid) == 0`, so the server's HEADERS land in `self._pending[sid]`, and then the same loop raises `h3 client: upload flow-control timeout` at `timeout_ms` and drops them. A server that answers 413 and withholds credit gives a `Response` on h2 and an exception on h3.

`head_ready(sid)` (`flare/http3/client.mojo:519`) after each chunk, breaking out to `Http3Download` the way the h2 path does, would close both halves. While you are in here: a `STOP_SENDING` arriving mid-upload is not observed on this path at all, for the `stream_reset` / `apply_stop_sending` mismatch I raised on #20.

flare/http/_client/stream_request.mojo
```diff
@@ -0,0 +1,245 @@
...
+        var next_chunk = source.next(cancel)
+        if not next_chunk:
+            break
+        var chunk = next_chunk.take()
+        if len(chunk) == 0:
+            continue
+        conn.send_data(sid, Span(chunk), False)
+        wire = conn.drain()
+        transport.write_all(Span(wire))
+        while conn.has_pending_body(sid):
```
ehsanmok, line 228: The connection is read only while there is a stashed body remainder, so the check below is really "early response noticed once the send window runs out". A peer that keeps WINDOW_UPDATEs ahead of us, or raises `SETTINGS_INITIAL_WINDOW_SIZE` in its SETTINGS, never makes `has_pending_body` true, and a 413 sent after the first megabyte sits in the socket buffer until the source is drained.

Your test covers the exhausted-window case well, RST_STREAM(NO_ERROR) variant included. What I would like is either a bounded poll of the transport in the outer loop between chunks, or the summary bullet and the `send_chunked` docstring narrowed to say the check happens when the upload blocks on credit.

flare/http/_client/stream_request.mojo
```diff
@@ -0,0 +1,245 @@
...
+            var n = transport.read(scratch.unsafe_ptr(), len(scratch))
+            if n == 0:
+                raise Error("HTTP/2 upload: EOF while waiting for flow control")
+            conn.feed(Span(scratch)[:n])
+            if conn.conn.goaway_sent:
+                raise Error("HTTP/2 upload: connection protocol error")
+            if conn.stream_error(sid):
+                raise Error("HTTP/2 upload: stream reset")
+            if conn.response_ready(sid):
+                conn.finish_upload(sid)
```
ehsanmok, line 240: `finish_upload` half-closes with an empty DATA plus END_STREAM (`flare/http2/client.mojo:626`) where RFC 9113 section 8.1 has the client send RST_STREAM with NO_ERROR after an early reply. Raised on #19, which is the layer `finish_upload` arrives in. Nothing to do here: the response is already in hand and the connection closes straight after.

tests/http/test_client_stream_upload_protocols.mojo
```diff
@@ -0,0 +1,187 @@
...
+struct Chunks(ChunkSource, Copyable, Movable):
+    var left: Int
+
+    def next(mut self, cancel: Cancel) raises -> Optional[List[UInt8]]:
+        if self.left == 0 or cancel.cancelled():
+            return None
+        self.left -= 1
+        var bytes = List[UInt8]()
+        bytes.resize(8192, 97)
+        return bytes^
```
ehsanmok, line 33: Could you add a variant of this source that raises on, say, its third `next`? It is the one wire-level behaviour in the PR with nothing covering it: on a mid-upload raise we must not emit `0\r\n\r\n` on h1 or the empty END_STREAM DATA on h2, or the server reads a truncated body as a complete one.

I traced it and it is correct today, but only through destructor ordering. The raise skips `flare/http/client.mojo:1218-1219` and `flare/http/_client/stream_request.mojo:244`, the owned `_H2Transport` drops, and `TcpStream.__deinit__` closes the socket. Nothing states that, so a refactor that parks the transport in a longer-lived owner would break it without failing a test.

flare/http/client.mojo
```diff
@@ -1096,119 +1108,116 @@ struct HttpClient(Movable):
...
-            NetworkError: On connection or I/O failure.
-        """
-        var u = Url.parse(self._resolve_url(url))
-
-        var host_header = u.host
-        if (u.scheme == "http" and u.port != 80) or (
-            u.scheme == "https" and u.port != 443
-        ):
-            host_header = host_header + ":" + String(Int(u.port))
+        """Convenience wrapper for an unknown-length streamed upload."""
```
ehsanmok, line 1111: This docstring replaced the one that spelled out the tradeoffs, and that old text was the only place the h1-only contract was written down. The four-argument overload is where the existing callers are, and for them this PR changes both the wire (TLS negotiates h2 now, cleartext honours `prefer_h2c`) and the headers (stored `Authorization`, the cookie jar, `Accept-Encoding: identity`). One line here saying the protocol is negotiated and that client auth and cookies apply would cover the API surface; the full list belongs in the commit message, which is currently just the subject line.

flare/http/client.mojo
```diff
@@ -1096,119 +1108,116 @@ struct HttpClient(Movable):
...
+                " supplied"
+            )
+        var url = self._resolve_url(req.url)
+        var method = req.method.upper()
+        var supplied = req.headers.copy()
+        var cookie = self._cookies.request_header()
+        if cookie.byte_length() > 0:
+            supplied.set("Cookie", cookie)
+        var headers = prepare_stream_headers(
+            supplied, -1, self._user_agent, self._auth_header
```
ehsanmok, line 1138: `prepare_stream_headers` already handles the known-length case (`flare/http/_client/stream_request.mojo:74-75` sets `Content-Length` when `body_size >= 0`), but this call hardcodes `-1` and the caller's own `Content-Length` is stripped at line 53 of the same file. So there is no way to stream a body whose size is known, which is the ordinary S3-style upload. Not for this PR. I am noting it because the plumbing is already there and the follow-up is one argument wide.

flare/http/client.mojo
```diff
@@ -1096,119 +1108,116 @@ struct HttpClient(Movable):
...
         )
-        var wb = wire.as_bytes()
-        stream.write_all(Span[UInt8, _](wb))
+        var response = self._upload_exchange(method, url, headers, source)
+        self._record_stream_head(url, response.headers)
+        var body = response.read_all()
```
ehsanmok, line 1142: `read_all()` has no ceiling, so the reply to a one-shot upload is buffered without bound while `read_all_limited` sits right next to it. Raised on #21, where that default lives; the existing buffered `send()` path has the same gap, so it is not a regression from your change and the follow-up is mine.

ehsanmok (changes requested):
Thanks for this - a streamed upload is the piece I most expected to go wrong, and the shape is right. Framing is generated by the wire driver, caller framing headers are stripped in one place, and the one-shot property falls out of simply not calling the retry and redirect machinery. I rebuilt the head locally and reran the focused suite before reviewing.

Two things I want addressed before merge, both on the "stop consuming the source when the server answers early" promise in the summary. On h3 it is not implemented at all, and the response the client has already parsed gets thrown away. On h2 it fires only while the send window happens to be exhausted, which is narrower than the bullet reads. Everything else below I am happy to carry as follow-ups.

On the rebase: this conflicts with `main` in `flare/tls/stream.mojo` and `tests/http/test_client_proxy.mojo`. Run `pixi run tests-gen` after rebasing so `tests/http/test_client_stream_upload_protocols.mojo` joins `tests/_agg`; see my note on #17 for what that runner does. I could not check the description's "focused streaming test suite passed after rebasing the complete stack onto current `main`" - the stack does not rebase cleanly today, so whatever resolution you ran that on is not in the PR.

## What I ran

Worktree at `02d1fd3`, Mojo 1.0.0 (ed45d567), macOS arm64.

| entry | result | wall | summary |
|---|---|---|---|
| `tests/http/test_client_stream_upload_protocols.mojo` | PASS | 24s | 5 tests run: 5 passed, 0 failed, 0 skipped |
| `tests/http/test_client_stream_upload.mojo` | PASS | 14s | 1 tests run: 1 passed, 0 failed, 0 skipped |
| `tests/http/test_chunked_request.mojo` | PASS | 12s | 8 tests run: 8 passed, 0 failed, 0 skipped |
| `tests/http/test_h3_live_dial.mojo` | PASS | 22s | 4 passed |

`mojo format` is clean on the 29 changed `.mojo` files. I did not run `git diff --check` separately.

The two pre-existing h1 suites pass unmodified, which is the check the approved spec asked for (`docs/superpowers/specs/2026-08-11-http-client-streaming-h2-h3-design.md`, testing plan). The h3 upload happy path is real coverage too. The file name says "live", but `test_h3_live_dial` binds a loopback `QuicListener` with the fixture CA, so the new 256 KiB `send_chunked` assertion runs everywhere the rest of the suite does.

## 1. h3 never looks for an early response, and discards one it already holds

`flare/http/client.mojo:1188-1197` pulls the source to exhaustion and only then builds the `Http3Download`. There is no `head_ready` check anywhere in the loop, so the h2 behaviour this PR is built around has no h3 counterpart.

On its own I would take that as a follow-up. What makes it a blocker is how it lands on top of the bounded credit wait from #20. `_send_request_wire` at `flare/http3/client.mojo:423-443` spins while `send_capacity(sid) == 0`, calling `poll_responses(100)` each time, which parses the server's HEADERS into `self._pending[sid]`. Then it raises `h3 client: upload flow-control timeout` once `timeout_ms` elapses and the parsed response goes with it. Same exchange, same server: a 413 on h2, an exception on h3.

The second half is `STOP_SENDING`: this loop keeps encoding DATA into a stream the peer has already asked it to abandon, because `stream_reset` tests only `STREAM_STATE_RESET_RECVD` while `apply_stop_sending` sets `STREAM_STATE_RESET_SENT`. Raised on #20, where that asymmetry lives; I mention it here because the upload loop is the path that reaches it.

Inline with what I think the smallest fix looks like.

## 2. On h2 the early-response check only fires while the send window is exhausted

`flare/http/_client/stream_request.mojo:228` reads the connection only inside `while conn.has_pending_body(sid)`. That is the case your test exercises, and the RST_STREAM(NO_ERROR) variant in it is a nice touch. But a peer that keeps WINDOW_UPDATEs ahead of the sender, or raises `SETTINGS_INITIAL_WINDOW_SIZE`, never makes `has_pending_body` true, the socket is never read, and a 413 sent after the first megabyte goes unseen until the source is drained. A multi-GB one-shot upload is exactly the case the feature exists for.

The summary bullet reads as a general guarantee with the exhausted-window case called out as the hard sub-case. What is implemented is the sub-case only. Either poll between chunks, or narrow the bullet and the `send_chunked` docstring to say the check happens when the upload blocks on credit. I do not mind which, as long as the two agree.

## 3. Existing `send_chunked` callers get a different wire and different headers, and the commit message is empty

The four-argument overload still compiles, so this is source-compatible, and the negotiation change is what the approved spec asked for. But several behaviours moved at once, and the unique commit for this layer has a subject line and nothing under it:

- TLS no longer pins ALPN to `http/1.1`, so an existing `https://` upload now goes out as h2 DATA (`_stream_transport` offers `["h2", "http/1.1"]` at `flare/http/client.mojo:1487-1496`). The ALPN change itself is raised on #17; what is new here is that existing `send_chunked` callers ride it.
- Cleartext honours `prefer_h2c` as prior knowledge (`flare/http/client.mojo:1497-1498`), so a `prefer_h2c=True` client changes wire too.
- The client's stored `Authorization` and the cookie jar are now attached, and `Set-Cookie` is recorded (`flare/http/client.mojo:1134-1141`). The old path sent neither. Stored auth also overrides a caller-supplied `Authorization` (`flare/http/_client/stream_request.mojo:71-73`), which your TLS-versus-cleartext test pair pins nicely, but it deserves writing down.
- `Accept-Encoding: identity` is forced and the reply goes through `_maybe_decompress`.
- Two new hard errors for these callers: a non-empty `Request.body` (`flare/http/client.mojo:1126`) and any `Expect` header (`flare/http/_client/stream_request.mojo:61-64`, which arrives with #21).

House rule here is that rationale lives in the commit message. Please put that list in the commit body, with the spec link, since the spec is what makes the ALPN change deliberate rather than a regression.

## 4. The one test I want in this PR

A `ChunkSource` that raises partway through. It is the only wire-level behaviour in the diff with no coverage, and the invariant is sharp: on a mid-upload raise the client must never emit the terminating `0\r\n\r\n` on h1 or the empty DATA with END_STREAM on h2, or the server reads a truncated body as a complete one.

I traced it and it is correct today, but only through destructor ordering. The raise skips `flare/http/client.mojo:1218-1219` and `flare/http/_client/stream_request.mojo:244`, the owned `_H2Transport` drops, `__deinit__` calls `Pool[TcpStream].free`, and `TcpStream.__deinit__` closes the socket. Nothing in the code says that ordering is load-bearing, and it is the kind of thing a later refactor breaks quietly.

- One-shot really is one-shot. `send_chunked` goes through neither `_send_with_retry` nor `_stream_once`'s redirect loop, `_stream_transport` always dials fresh and never consults `_tls_pool`, h1 sends `Connection: close`, and `HttpStreamResponse.close` frees the connection instead of releasing it to a pool. A 3xx comes back to the caller as a plain `Response`, which is the right answer for a body that cannot be replayed.
- No concurrent exposure: the reply is fully buffered into a `Response` before `send_chunked` returns, so there is never a live upload and a live reader at the same time.
- Caller framing is genuinely gone. `prepare_stream_headers` strips `Content-Length`, `Transfer-Encoding`, `Connection` and its nominated tokens, `Keep-Alive`, `Proxy-Connection`, `Upgrade` and `HTTP2-Settings` before any wire sees them (`flare/http/_client/stream_request.mojo:45-60`), and a CR or LF in any surviving value raises.
- `_build_h2_request_headers(headers, "", "")` in `upload_h2` looked wrong to me at first read. It is fine: `prepare_stream_headers` has already set `User-Agent`, so the empty fallback is never appended.
- Leaving `Expect: 100-continue` unsupported is the right scope for this PR. I am noting it only because a rejected 5 GB upload is precisely what 100-continue exists for, so it belongs on the list below rather than being forgotten.

## What I'll take

None of these need to hold the PR up. I'll open issues unless you'd rather fold them in.

- A known-length streamed upload. `prepare_stream_headers` already honours `body_size >= 0` (`flare/http/_client/stream_request.mojo:74-75`) but `send_chunked` hardcodes `-1` and the caller's own `Content-Length` is stripped, so a body of known size cannot be streamed at all.
- `Expect: 100-continue` on the upload path.
- `ChunkSource` trailers. The spec deferred these and I am content to leave them deferred, but the reply side now does surface trailers, so the asymmetry is worth recording.
- A ceiling on the buffered reply at `flare/http/client.mojo:1142`. Raised on #21, where the unbounded `read_all` default lives.
- The ALPN-degrades-to-h1-over-TLS fallback test the spec's testing plan asked for, and a test that a 3xx comes back to the caller.

From you: the h3 early-response check in item 1, the narrowing or the poll in item 2, the commit body in item 3, and the raising `ChunkSource` test in item 4.

ehsanmok:

here.

PR #25, opened 2026-09-09
uds(ffi): abstract-namespace sockaddr encoding (@name convention)

zharia:
Additive: fill_sockaddr_un encodes leading-@ names as Linux abstract sockets (sun_path[0]=0x00 + name, no trailing NUL, addrlen = 2 + 1 + len(name)); pathname branch byte-identical; abstract raise on macOS/BSD layouts (sun_path[0] is sun_len). Called out at the HyrxMQ consumers' request (HyrxMQ 0011); no test or behavior change for existing path sockets.


zharia:
Implemented Linux abstract domain sockets. Enables establishing unix domain socket connections without touching the file system

PR #29, opened 2026-09-18
perf(grpc): let a server-streaming handler decline compression

### LINKED ISSUE #27: gRPC server-streaming compresses every response the client says it accepts
Branch: [`millfolio:perf/grpc-decline-compression`](https://github.com/millfolio/flare/tree/perf/grpc-decline-compression) · [compare](https://github.com/ehsanmok/flare/compare/main...millfolio:flare:perf/grpc-decline-compression) · one commit on `main` (a35e1f5)

## What happens

`_negotiate_response_encoding` treats `grpc-accept-encoding: gzip` as an
instruction rather than as permission, so every server-streaming response past
`GRPC_COMPRESS_MIN_BYTES` is deflated. Every gRPC client advertises gzip — it
is in grpc-core's default headers — so in practice every large response is
compressed and a handler has no way to decline.

For a payload that is already compact, that is pure cost. I hit it building an
Arrow Flight server on flare: Arrow record batches are laid out so a consumer
can cast the buffers where they land, and they were being gzipped on the way
out at about 25 MB/s.

## Measurement

Serving one column of a 79,478,796-row Iceberg table, 28 MiB per endpoint,
macOS on an M4. `sample` of the server under load:

```
715  flare_compress_gzip  (in libflare_zlib.so)
       715 deflate  (in libz.1.3.2.dylib)
         650 deflate_slow
           650 longest_match
```

715 of ~1000 samples. The handler that produced those 28 MiB took 16 ms; the
client waited 1236 ms for them. Declining compression on that one call:

| | per endpoint |
|---|---|
| before | 1236 ms |
| after | 385 ms |

## What the branch does

Not a change of default — that is a behaviour decision and yours to make. It
adds `GrpcServerStreamReply.compress`, defaulting to `True`, and honours it
in both places that negotiate an encoding, so a handler that knows what it is
sending can say so. Existing handlers are unaffected.

Worth noting for the default question: grpc-go, grpc-java and grpc-c++ all
default to identity for exactly this reason — accepting an encoding is not
asking for one, and the sender chooses.

## Tests

`test_grpc_server_stream` (3), `test_grpc_compression` (5),
`test_grpc_streaming` (6), `test_grpc` and `test_grpc_proto` (6) all pass on
the branch.

Happy to turn this into a PR when the API shape looks right to you, or to drop
it if you would rather change the default instead.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GcgGYzWPPi9uQdxj7KybBj

winding-lines:
Closes #27.

`_negotiate_response_encoding` treats `grpc-accept-encoding: gzip` as an
instruction rather than as permission, so every server-streaming response past
`GRPC_COMPRESS_MIN_BYTES` is deflated and a handler has no way to say no.
Every gRPC client advertises gzip, so in practice every large response is
compressed.

This does **not** change that default — that is a behaviour decision and yours
to make. It adds `GrpcServerStreamReply.compress`, defaulting to `True`, and
honours it in both places that negotiate an encoding. Existing handlers are
unaffected; a handler that knows its payload is already compact can now opt
out.

### Why it matters

Serving one column of a 79.5M-row Iceberg table over Arrow Flight, 28 MiB per
endpoint. `sample` of the server under load:

```
715  flare_compress_gzip  (in libflare_zlib.so)
       715 deflate  (in libz.1.3.2.dylib)
         650 deflate_slow
           650 longest_match
```

715 of ~1000 samples, compressing Arrow record batches at about 25 MB/s. The
handler that produced those bytes took 16 ms; the client waited 1236 ms.
Declining compression took that call to 385 ms.

Arrow's whole argument is that the consumer casts the buffers where they land,
so deflating them on the way out gives that up twice over — but the same is
true of any payload a handler has already packed.

Worth noting for the default question, which I have deliberately left alone:
grpc-go, grpc-java and grpc-c++ all default to identity, on the reasoning that
accepting an encoding is not asking for one and the sender chooses.

### Tests

`test_grpc_server_stream` (3), `test_grpc_compression` (5),
`test_grpc_streaming` (6) and `test_grpc_proto` (6) pass on this branch,
re-run against `main` at a35e1f5.

The issue has the full profile and a before/after table.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GcgGYzWPPi9uQdxj7KybBj


ehsanmok (approved):
LGTM! thanks

PR #30, opened 2026-09-18
perf(http2): stop copying a response body per WINDOW_UPDATE, and per frame

### LINKED ISSUE #28: HTTP/2: a parked response body is re-copied on every WINDOW_UPDATE (quadratic in body size)
Branch: [`millfolio:perf/bulk-copies`](https://github.com/millfolio/flare/tree/perf/bulk-copies) · [compare](https://github.com/ehsanmok/flare/compare/main...millfolio:flare:perf/bulk-copies) · one commit on `main` (a35e1f5)

## What happens

A response too large for the peer's send window is parked and re-pumped when
a WINDOW_UPDATE opens it. Both pump paths hand the framing layer the **whole**
remainder each time and let the window decide how much of it to take:

```mojo
# Http2Connection.pump_pending
var body = self.pending_body[sid].copy()      # the whole parked body
var pos = self.pending_pos[sid]
var rest = List[UInt8](capacity=len(body) - pos)
for k in range(pos, len(body)):               # and then the whole tail,
    rest.append(body[k])                      # one byte at a time
var n = self.queue_stream_data(sid, Span[UInt8, _](rest))
```

`Http2ConnHandle._pump_one` does the same for a streaming response's stash.

A WINDOW_UPDATE is an inbound frame and `pump_pending` runs after every
inbound frame, so a body large enough to park is copied twice per window's
worth of progress. That is **quadratic in the body**: at a 64 KiB window a
28 MiB response copies about 6 GB to send 28 MB, most of it one byte per
iteration. The work grows with the square of the response while the wire time
grows linearly, which is why it does not show up on the bodies the tests
cover — it only appears once a response is large enough to park repeatedly.

## Measurement

An Arrow Flight server built on flare, serving one column of a 79,478,796-row
Iceberg table (28 MiB per endpoint), macOS on an M4. The handler produces
those bytes in 16 ms; the numbers below are what the client waits:

| | per endpoint | whole column, 24 endpoints |
|---|---|---|
| before | 152 ms | 800 ms |
| after | 58 ms | 391 ms |

With the profile flat after the fix — the remaining time is zstd-decompressing
Parquet, which is the work.

## What the branch does

Both pumps ask what the window will take (a small `_send_budget` helper,
which is the bound `queue_stream_data` already applies) and copy exactly that
much, once.

It also makes five linear copies on the same path one `memcpy` each, since
they are the same mistake at a smaller scale and were measurable behind the
quadratic: `queue_stream_data` copying each DATA payload and then appending
the encoded frame to the outbox a byte at a time, `encode_frame` copying the
payload a third time, `ProtoWriter.write_bytes` appending a length-delimited
field byte by byte (invisible on a string field, the whole payload on a gRPC
message body), `ProtoReader` taking its copy the same way, and the three
places that shovel a drain into `write_buf`. `drain` hands its buffer over
rather than copying it.

Happy to split the quadratic from the memcpys if you would rather review them
separately — they are independent.

## Tests

`test_h2_conn_handle` (3), `test_h2_server` (9), `test_h2_frame` (9),
`test_h2_state` (14), `test_h2_config` (16), `test_stream_slab` (10),
`test_h2_server_handler` (5), `test_grpc_proto` (6), `test_grpc_streaming`
(6), `test_grpc_server_stream` (3), `test_grpc_compression` (5) and
`test_backpressure` pass on the branch. `test_stream_coalescing` fails
identically on unmodified `main` on this machine, so it is not from this
change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GcgGYzWPPi9uQdxj7KybBj

winding-lines:
Closes #28.

A response too large for the peer's send window is parked and re-pumped when a
WINDOW_UPDATE opens it. Both pump paths hand the framing layer the **entire**
remainder each time and let the window decide how much to take —
`Http2Connection.pump_pending` copies the parked body and then copies its tail
a byte at a time, and `Http2ConnHandle._pump_one` does the same for a
streaming response's stash.

That is quadratic in the body. At a 64 KiB window a 28 MiB response copies
about **6 GB to send 28 MB**, most of it one byte per iteration: the work grows
with the square of the response while the wire time grows linearly. It is
invisible on the small bodies the tests cover, which is why it survived.

Both now ask what the window will take — a small `_send_budget` helper, which
is the bound `queue_stream_data` already applies — and copy exactly that much,
once.

The same shape at a smaller scale, linear but on every byte, in five more
places on that path, all now one `memcpy` each:

| where | what it did |
|---|---|
| `queue_stream_data` | copied each DATA payload out of the response, then appended the encoded frame to the outbox a byte at a time |
| `encode_frame` | copied the payload into the wire bytes a third time |
| `ProtoWriter.write_bytes` | appended a length-delimited field byte by byte — invisible on a string field, the whole payload on a gRPC message body |
| `ProtoReader.__init__` | took its copy the same way |
| `Http2ConnHandle` (×3) | shovelled a drain into `write_buf` a byte at a time |

`drain` now hands its buffer over instead of copying it.

### Measured

An Arrow Flight server built on flare, serving one column of a 79.5M-row
Iceberg table, 28 MiB per endpoint. The handler produces those bytes in 16 ms;
these are what the client waits:

| | per endpoint | whole column, 24 endpoints |
|---|---|---|
| before | 152 ms | 800 ms |
| after | 58 ms | 391 ms |

After it, the server's profile is flat — what remains is zstd-decompressing
Parquet, which is the work.

### Tests

`test_h2_conn_handle` (3), `test_h2_server` (9), `test_h2_frame` (9),
`test_h2_state` (14), `test_grpc_streaming` (6) and `test_backpressure` pass,
re-run against `main` at a35e1f5. The full `pixi run tests` result is on the
issue: five suites fail and all five fail identically on unmodified `main`
here (`test_stream_coalescing`, `agg_tls`, `test_block_in_pool`, and the
`cert_reload` / `mtls` examples).

Happy to split the quadratic from the memcpys into two PRs if you would rather
review them separately — they are independent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GcgGYzWPPi9uQdxj7KybBj


flare/grpc/proto.mojo
```diff
@@ -158,10 +158,25 @@ struct ProtoWriter(Copyable):
...
         self._tag(field, WIRE_LEN)
         self._raw_varint(UInt64(len(value)))
-        for i in range(len(value)):
-            self.buf.append(value[i])
+        var n = len(value)
+        if n == 0:
+            return
+        var base = len(self.buf)
+        self.buf.resize(unsafe_uninit_length=base + n)
+        unsafe_memcpy(
```
ehsanmok: Why unsafe? should be doable with the same api

flare/http/_h2_conn_handle.mojo
```diff
@@ -421,9 +422,28 @@ struct Http2ConnHandle(Movable):
...
+
+            # connection, and for a streaming response it runs once
+
+            # per window's worth of body.
+
+            var base = len(self.write_buf)
+
+            self.write_buf.resize(unsafe_uninit_length=base + len(ack))
+
+            unsafe_memcpy(
```
ehsanmok: similar concern and looks like repeating the logic in multiple places?

flare/http/_h2_conn_handle.mojo
```diff
@@ -553,9 +573,28 @@ struct Http2ConnHandle(Movable):
...
+
+            # connection, and for a streaming response it runs once
+
+            # per window's worth of body.
+
+            var base = len(self.write_buf)
+
+            self.write_buf.resize(unsafe_uninit_length=base + len(out))
+
+            unsafe_memcpy(
```
ehsanmok: same

ehsanmok (changes requested):
Thanks! concern over the use of unsafe_memcopy

flare/http/_h2_conn_handle.mojo
winding-lines: sorry about the repeated and bad pattern

flare/grpc/proto.mojo
winding-lines: You were right — it is doable with the same API and it costs nothing. Every
copy in this PR is `List.extend(Span)` now; `unsafe_memcpy` is gone from the
diff entirely.

I measured before swapping, since the whole PR is a performance claim. Copying
28 MiB into a `List[UInt8]`, p50 of five runs:

| | |
|---|---|
| append, a byte at a time | 14.967 ms |
| `extend(Span)` | 0.625 ms |
| `resize(unsafe_uninit_length=)` + `unsafe_memcpy` | 0.590 ms |

6% between the two spellings that matter, against 24× for the one being
removed. The end-to-end number is unchanged too: the Flight server that
prompted this still serves its column in 384 ms, the same as with the unsafe
version.

flare/http/_h2_conn_handle.mojo
winding-lines: Factored. `Http2Connection.queue_parked_body(sid, body, pos)` now owns the
whole "ask the window what it will take, copy exactly that, frame it" step,
and both pump paths call it — the buffered one in `pump_pending` and the
streaming one here in `_pump_one`. Neither spells out budget, clamp and copy
any more.

One wrinkle worth flagging, because it is the reason `pump_pending` reads
slightly oddly: the parked body lives in `self.pending_body`, and the helper
takes `self` mutably, so passing a span of it would alias. It pops the body
out of the map and puts it back if the stream is not finished. A `List` move
is a pointer transfer, so that costs nothing per pump.

flare/http/_h2_conn_handle.mojo
winding-lines: This one turned out not to need a helper: with `extend` the whole site is a
single line — `self.write_buf.extend(Span(out))` — so the three copies of it
are three identical one-liners rather than three copies of a block. Happy to
pull them into a small method if you would still rather have one.

winding-lines:
Pushed. Both concerns addressed:

**`unsafe_memcpy` is gone.** Every copy is `List.extend(Span)`. I measured
first rather than assume, since the PR is a performance claim — 28 MiB into a
`List[UInt8]`, p50 of five: byte-at-a-time 14.967 ms, `extend(Span)`
0.625 ms, `resize` + `unsafe_memcpy` 0.590 ms. The safe spelling is within 6%
of the unsafe one and 24× faster than what is being removed, so there was
never a reason to reach past it.

**The duplication is in one place.** `Http2Connection.queue_parked_body` holds
the bounded-copy step and both pump paths call it. The three
drain-into-`write_buf` sites are one `extend` each now, so nothing is left to
factor there.

The end-to-end number is unchanged in the Arrow Flight server that prompted this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GcgGYzWPPi9uQdxj7KybBj

ehsanmok (approved):

